PR #28, opened 2022-12-27
Exclude inflate tests from the published package

Shnatsel:


etemesi254:
Thanks for catching this

PR #39, opened 2022-12-31
Don't waste time initializing memory that we're going to overwrite

Shnatsel:
This requests uninit memory from the OS and writes to it directly instead of asking the OS to zero the memory first.

`.collect()` is handy like that.


etemesi254:
Pretty nice trick, wasn't aware.

Thanks

PR #48, opened 2023-01-01
Expand roundtrip fuzzing of PNG to funky bit depths

Shnatsel:
Finds mismatches with the original length of the input. Probably some exotic bit depth isn't properly supported.


etemesi254:
Yea, currently support for 8 bit and 16 bit is fully complete. The last step is bits less than 8 and the png decoder would be complete and conforming.

So this may still cause crashes

PR #49, opened 2023-01-02
update

etemesi254:


.github/workflows/fuzz_deflate.yml
```diff
@@ -19,11 +19,13 @@ jobs:
...
+        toolchain: nightly
+        override: true
     - name: Install
       run: cargo install cargo-fuzz
     - name: Fuzz deflate
       run: cargo +nightly fuzz run --fuzz-dir zune-inflate/fuzz roundtrip -- -runs=10000
     - name: Fuzz zlib
       run: cargo +nightly fuzz run --fuzz-dir zune-inflate/fuzz roundtrip_zlib -- -runs=10000
+    - name: Normal Fuzz testing.
+      run: cargo +nightly fuzz run --fuzz-dir zune-inflate/fuzz decode_buffer  -- -runs=10000
```
Shnatsel, line 31: 10,000 runs of the `roundtrip` target complete in 4 seconds on my machine, and doesn't provide good coverage.

Perhaps bump it to 100,000? That takes 20 seconds on my machine and has 2x better coverage, close to what I got after running the fuzzer for hours.

Alternatively you can commit the fuzz corpus to the repo so that the fuzzer isn't starting from scratch every time.

etemesi254: Just fixed this currently, for jpeg and png. They now run with a corpus.

For inflate/deflate/gzip none of them have a checksum hence this still hits most of the paths.

Currently its 200,000 though.

We finally have CI fuzz testing for png and jpg,ran once every day, jpg doesn't crash, png does, which is expected.

But id say its one step closer to production.

Also CI timing is limited, I mistakenly used 360 mins previously because i confused `--runs` with `-runs` and the fuzzer ran the whole night.

Though I'm setting up CI in a way that fuzzers only kick in when a change in that directory occurs and not a change in the whole project.

PR #62, opened 2023-01-27
Less bounds checks in RLE fill with a single byte

Shnatsel:
Do a bit more math up front in exchange for not having to perform two bounds checks per loop iteration on long RLE copies, and do nothing but copies in the loop.

I imagine PNGs filled with a single color should benefit, as well as screenshots or anything with large flat fill. Other cases shouldn't regress much, but I haven't actually checked, my machine is too noisy for benchmarks.

### DIFF
--- a/zune-inflate/src/decoder.rs
+++ b/zune-inflate/src/decoder.rs
@@ -18,7 +18,7 @@ use crate::gzip_constants::{
     GZIP_CM_DEFLATE, GZIP_FCOMMENT, GZIP_FEXTRA, GZIP_FHCRC, GZIP_FNAME, GZIP_FOOTER_SIZE,
     GZIP_FRESERVED, GZIP_ID1, GZIP_ID2
 };
-use crate::utils::{copy_rep_matches, fixed_copy, fixed_copy_within, make_decode_table_entry};
+use crate::utils::{copy_rep_matches, fixed_copy_within, make_decode_table_entry};
 
 struct DeflateHeaderTables
 {
--- a/zune-inflate/src/decoder.rs
+++ b/zune-inflate/src/decoder.rs
@@ -958,8 +958,8 @@ impl<'a> DeflateDecoder<'a>
                         {
                             // RLE match, repeat the byte in batches of 8
                             const COPY_SIZE: usize = 8;
-                            let rep_num = u64::from(out_block[src_offset]) * 0x0101010101010101;
-                            let rep_byte = &rep_num.to_ne_bytes();
+                            let byte_to_repeat = out_block[src_offset];
+                            let slice_to_repeat = [byte_to_repeat; COPY_SIZE];
                             // compute much should be filled with this value
                             let fill_length = usize::saturating_sub(dest_offset, current_position);
                             // round the fill length up to increments of COPY_SIZE
@@ -968,7 +968,7 @@ impl<'a> DeflateDecoder<'a>
                             let fill_area = &mut out_block[current_position..][..sloppy_fill_length];
                             // perform the actual fill
                             fill_area.chunks_exact_mut(COPY_SIZE).for_each(|chunk| {
-                                chunk.copy_from_slice(rep_byte);
+                                chunk.copy_from_slice(&slice_to_repeat);
                             });
                         }
                         else if offset <= FASTCOPY_BYTES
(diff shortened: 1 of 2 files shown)


Shnatsel:
Also, I'm not sure why it's 8 instead of the `FASTCOPY_BYTES` like the rest of the copies, but I didn't change that. 

With SIMD there are 128-bit registers and loads/stores which potentially could be twice faster than this 64-bit copying.

Shnatsel:
I tried 128-bit sized copies and that made no difference whatsoever on my x86 machine. Neither does this, really.

Shnatsel:
I also have a slightly more idiomatic version that is entirely parametrized on one constant:
https://github.com/Shnatsel/zune-image/tree/rusty-rep-byte
Also no significant performance difference on my machine.

Shnatsel:
Benchmarking with the following code is far more successful than running a full PNG to PPM conversion:

```rust
fn main() {
    use zune_inflate::DeflateDecoder;
    let path = std::env::args_os().nth(1).unwrap();
    let input = std::fs::read(&path).unwrap();
    let mut decoder = DeflateDecoder::new(&input);
    
    let decompressed = decoder.decode_zlib().unwrap();
    println!("{}", decompressed.last().unwrap());
}
```

With this code it seems that the zero-filled file is processed 10% faster but the enwiki file seems to be unchanged or 1% or even 2% slower. 

This makes sense for the `enwiki` file - coverage shows that the loop is usually only entered once (2100 hits for RLE decode and only 2800 hits for the loop body), so the additional math we do up front hurts instead of helping.

I wonder if running a single cycle unconditionally and only doing the extra math and looping if the repeated sequence is longer than FASTCOPY_BYTES would be beneficial. Would you be OK with that approach? It does add extra complexity to an already complex code.

etemesi254:
> Also, I'm not sure why it's 8 instead of the `FASTCOPY_BYTES` like the rest of the copies, but I didn't change that.
> 
> With SIMD there are 128-bit registers and loads/stores which potentially could be twice faster than this 64-bit copying.

I was a bit pendantic there, SSE loads have a latency of 4-12 https://uops.info/html-instr/MOVDQU_M128_XMM.html#WOL

The other normal loads have a latency of 0-5 https://uops.info/html-instr/MOV_R64_M64.html.

Sadly in that loop it's not always about large copies, longer latencies puts more pressure on the memory unit, and CPU, so it may not always be faster.

etemesi254:
> With this code it seems that the zero-filled file is processed 10% faster but the enwiki file seems to be unchanged or 1% or even 2% slower.

What zero filled file?

Also btw the easiest way to copy rle matches  with offset 1 is the good old `memset`, i.e the code should be 

```Rust
let byte= out_block[src_offset];
out_block[current_position..dest_offset].fill(byte);
```

I didn't like this because since it will always call `memset` which may mess up with all cpu internal data structures and memset doesn't know about sloppy bytes so it can't use it to it's advantage.

 But  let benchmarks speak for themselves
 
|file| opt|numbers|
|-----|----|--------------|
|png-zlib|0|141.18|
|gzip-tokio|0|255|
|gzip-image-rs|0|78.027|
|png-zlib|1|140.96|
|gzip-tokio|1|263.04|
|gzip-image-rs|1|78.517|
|png-zlib|2|139.27|
|gzip-tokio|2|258.00|
|gzip-image-rs|2|76.93|

Opt 1 is the memset one above
Opt 2 is the one currently present in the loop
Opt 3 is the proposed change

Command is `RUSTFLAGS='-C target-cpu=native' cargo bench "zune"` inside `zune-inflate` directory.

Shnatsel:
Who knew that the most idiomatic solution is also the fastest one :sweat_smile: 

Although when comparing just opt 0 (memset) and opt 2 (this PR), I see in your benchmarks that opt 2 is faster in 2 cases out of 3 still. Perhaps using this PR but with one unconditional copy up front is worth a shot after all.

FWIW I was benchmarking without `RUSTFLAGS='-C target-cpu=native'`

etemesi254:
> FWIW I was benchmarking without `RUSTFLAGS='-C target-cpu=native'`
I just do it like that because I started like that, makes it easy to quantify speeds, i.e I know on my machine enwiki is at 162 mb/s, so if someone suggests a change which promises speed and I benchmark, I know if it's faster on my computer than 162mb/s, it's faster.

> Although when comparing just opt 0 (memset) and opt 2 (this PR), I see in your benchmarks that opt 2 is faster in 2 cases out of 3 still. Perhaps using this PR but with one unconditional copy up front is worth a shot after all.

I think its only faster in 1 out of 3, and even so opt 2 still beats it there

Shnatsel:
Well, the differences are so small that my machine has noise far stronger than this. So I think the best course of action is to just use `fill` and not complicate the code any more than necessary.

PR #74, opened 2023-02-01
PNG: Fuzz with bit depths both 8 and 16

Shnatsel:
Lower bit depths fail to roundtrip, I'm not sure why. The same works for `image-png` for all bit depths: https://github.com/image-rs/image-png/pull/376

So I've limited fuzzing to bit depths 8 and 16 for now, and added a TODO to expand it to other bit depths later.


etemesi254:
It's noted that we currently expand bit depths lower than 8 to 8, so that may cause faliure

PR #98, opened 2023-03-09
impl Error for DecodeErrors

Shnatsel:
Makes integration with `image` easier, among other things

### DIFF
--- a/zune-jpeg/src/errors.rs
+++ b/zune-jpeg/src/errors.rs
@@ -41,6 +41,9 @@ pub enum DecodeErrors
     /// Large image dimensions(Corrupted data)?
     LargeDimensions(usize)
 }
+
+impl std::error::Error for DecodeErrors {}
+
 impl From<&'static str> for DecodeErrors
 {
     fn from(data: &'static str) -> Self
--- a/zune-jpeg/src/errors.rs
+++ b/zune-jpeg/src/errors.rs
@@ -42,6 +42,7 @@ pub enum DecodeErrors
     LargeDimensions(usize)
 }
 
+#[cfg(feature = "std")]
 impl std::error::Error for DecodeErrors {}
 
 impl From<&'static str> for DecodeErrors


etemesi254:
It should be  covered with cfg std, to ensure it doesn't compile without std


Thanks


To make it even easier, I'm working on a perfomant  generic way to implement `Read ` so that we don't need to do the whole buffer into memory but haven't yet settled on a good way to express it, but i feel it should at least work by the time we're announcing zune-png

Shnatsel:
Oh! I literally just pushed initial integration into `image` that just buffers the whole thing in memory :sweat_smile: 

https://github.com/image-rs/image/compare/master...Shnatsel:image:zune-jpeg?expand=1

Shnatsel:
Added the feature gate on `std`

etemesi254:
> Oh! I literally just pushed initial integration into image that just buffers the whole thing in memory sweat_smile

The problem is a hard one to solve though,  wold appreciate suggestions, I'm trying to avoid double buffering situations for cases like where the object is already in memory, or where it's wrapped in a buffereadreader

Shnatsel:
That sounds like you need _specialization_ - a generic implementation for any `Read` but a specialized one for `BufRead`. That is sadly a nightly-only feature, but there are some hacks available on stable that make it work with some limitations.

etemesi254:
What hacks?

Shnatsel:
I don't remember the details. I've tried googling but it just shows a lot of incomplete approaches. 

So I've asked on the community discord [here](https://discord.com/channels/273534239310479360/592856094527848449/1083508178957914307) ([invite link](https://discord.gg/rust-lang-community)) and the answer was "no way unless the user goes through a macro, in which case you can use deref specialization or similar"

We can just make an unbuffered API requiring `BufRead`, and another buffered one for a generic `Read`. That means `image` will use the lowest common denominator and do extra buffering, but oh well :shrug: 

`Read` also kinda sucks in that it can't read into uninitialized buffers. At least the replacement is being actively worked on, unlike specialization.

fintelia:
`BufRead` is rather frustrating as a trait because it seems like it should let you get the a slice of the next N bytes from the reader, but it actually doesn't. The only method you have is `fill_buf` which contrary to the name only fills the buffer if it is empty. So if there's 1-byte left in the buffer your decoder has to be able to handle it. But the whole reason you wanted `BufRead` is to avoid having to handle a trickle of single-byte reads!

etemesi254:
Ideally I think the best API should be that the byte reader calls `read()` like a polling interface to ask for more bytes.

The problem then becomes byte boundaries, I.e when you need more bytes than what the buffer currently has unconsumed.

And also how to make that performant

etemesi254:
I tried it. 

And the number of syscalls I see happening does not make it worth it.

And it generates a lot of code, see https://godbolt.org/z/GonqbW68x

I'm afraid I'll have to stick with full memory buffering

Issue #12, opened 2022-12-13
Incorrect decoding of some real-world JPEG images


Shnatsel:
At a quick glance, there seem to be at least two issues going on:

1. Quality degradation / pixelation on some greyscale or mostly-greyscale images
2. Incorrect recovery after RST marker in damaged images, so that everything following the RST marker is misaligned slightly
3. There are also some small images that are completely garbled, but that might be due to issue 1, not sure

Samples for reproducing the issues: [zune-divergences-3.tar.gz](https://github.com/etemesi254/zune-image/files/10213871/zune-divergences-3.tar.gz)

Tested on commit 5877c2f2bf984f9f31b4883a1a4df628e5e0ae9b

etemesi254:
This one might take some time.

Shnatsel:
I don't really care what happens to truncated images. No reason to try to fix them if this slows down decoding of valid images!

etemesi254:
Issue no 3 should be fixed, took long because I was barking up the wrong tree, focusing on issue no 2 and 1 now.

etemesi254:
1 is fixed, really appreciated this one, helped me find a nasty bug. Will update with benchmarks soon

etemesi254:
Updated benchmarks(https://github.com/etemesi254/zune-image/blob/dev/zune-jpeg/Benches.md), we are slightly faster than mozjpeg which wraps  libjpeg-turbo but would be nice if you could test to ensure we're also correct now.  I'll be looking into the RST issue soon but I do feel we shouldn't be having huge divergences from imagemagick.

Thank you for your work.

Feel free to close this if there are no further issues

Shnatsel:
Unfortunately this has regressed many images. Here's one example, but they all seem to show the same artifacts:

[philosoguar](https://user-images.githubusercontent.com/291257/210149396-22834c32-15dd-4c9b-8b5f-71c796735822.jpg)

This is my own drawing, I'm licensing it under [CC-BY 4.0](https://creativecommons.org/licenses/by/4.0/) so you could include it in tests.

etemesi254:
Hi, could you check now under the dev branch.

Should hopefully be fixed.

I'm trying to figure out a way to ensure these regressions do not happen
again

Shnatsel:
I can confirm this is fixed on the `dev` branch.

Shnatsel:
[...]

PR #105, opened 2023-03-14
Fix possible off by one error in ac refinement

vstroebel:
This is a possible fix for the Lion Image in #12 .

In cases  where the while loop `'advance_nonzero` is taken, the variable k (Coefficient index) might be +1 off after the loop, which results in the modification of the wrong coefficient after the loop.
In most cases this only results in minor differences in the decoded image, but in case of k=64 it wraps around to 0 and changes the DC value. This should be the root cause for the strange blocks in the Lion Image.


etemesi254:
Thank you for this.

Let me rebase and see if it passes image tests and if it does we can make a release

(The CI was failing at a totally unrelated part)

PR #116, opened 2023-04-02
Remove bounds checks from bit expansion

Shnatsel:
Opening this as a draft to get feedback on this before the test run.

I added these to honor the existing APIs:

```rust
    let input = &input[in_offset..];
    let out = &mut out[..width * out_n];
```

~~but I'm not sure if it's best placed here, or if the caller should figure out the appropriate slices instead.~~

but `in_offset` is always 0 and the other slice is always greater than `width * out_n` so this doesn't actually matter and these arguments can be removed.

### DIFF
--- a/zune-png/src/decoder.rs
+++ b/zune-png/src/decoder.rs
@@ -980,7 +980,6 @@ impl<'a> PngDecoder<'a>
                         expand_bits_to_byte(
                             width,
                             usize::from(info.depth),
-                            0,
                             n_components,
                             self.seen_ptle,
                             to_filter_row,
@@ -999,7 +998,6 @@ impl<'a> PngDecoder<'a>
                         expand_bits_to_byte(
                             width,
                             usize::from(info.depth),
-                            0,
                             n_components,
                             self.seen_ptle,
                             &self.previous_stride,
@@ -1089,7 +1087,6 @@ impl<'a> PngDecoder<'a>
                         expand_bits_to_byte(
                             width,
                             usize::from(info.depth),
-                            0,
                             n_components,
                             self.seen_ptle,
                             to_filter_row,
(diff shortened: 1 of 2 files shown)


Shnatsel:
My machine is too noisy for benchmarks, but `cargo asm` shows 2 panic landing pads in the function instead of the previous 10.

Shnatsel:
Hmm, this broke interlaced 1bpp decoding and I don't understand why.

etemesi254:
```Rust
 // handle the remainder at the end where the output is less than 8 bytes long
        if let Some(in_val) = in_iter.next() {
            let remainder_iter = out_iter.into_remainder().iter_mut();
            remainder_iter.enumerate().for_each(|(pos, out_val)| {
                let shift = (7_usize).wrapping_sub(pos);
                *out_val = scale * ((in_val >> shift) & 0x01);
            });
        }
```
If current is < 8 px means the main chunks loop will never iterate and the remainder loop which will iterate should start at position zero of `in_iter`

But this 

```Rust
  (&mut in_iter).zip(&mut out_iter).for_each(|(in_val, out_vals)| 
``` 

will advance in_iter once(implicitly call next(, find out that the `zip` is none and exit out, 

but remember we moved  that one step,so the remainder will use one step further.

The failing test-case looked like this

```
input = [240,0,0,0];
width =4;
```

`in_val` pointed to 0 for your remainder loop instead of 240.


Interestingly if you switch loop order, the test cases pass as we only advance in_iter if we know out will be filled

```Rust
 (&mut out_iter).zip(&mut in_iter).for_each(|(out_vals, in_val)
      
```

etemesi254:
> other slice is always greater than `width * out_n`

Let this one remain, since we are using zips, it has the effect of ensuring we stop at exactly the right place or we
can as you suggested, figure out the right slice before calling it, (that's better).


> but in_offset is always 0

Should be removed

Shnatsel:
Oh, that's a tricky failure mode! Great catch!

I've also removed `in_offset` as you instructed. Let me run another test on my corpus.

Shnatsel:
CI seems to be broken by unrelated changes. I'll sync to make it green once it's fixed in the `dev` branch.

Shnatsel:
The test on my corpus is clean, marking ready for review.

PR #119, opened 2023-04-02
Call out APNG in zune-png README

Shnatsel:


etemesi254:
There are plans to support it, but want to do it correctly, so still investigating.

The main problem is something like APNG-> jpeg and what it should produce and how to handle frame offsets and blend modes.

PR #124, opened 2023-04-24
decoder colorspace override

PartTimer1996:
Useful for the edge case of manually creating jpegs and providing adobe headers - this will assign the file to be CMYK when RGB is actually the valid colorspace. In this case the manual override allows me to use the package when the metadata is being interpreted incorrectly. This is not a perfect fix but good for my use case with time tight.


etemesi254:
Hi, thanks for the PR, 

Do you mean the adobe headers and the jpeg exist separately or your adobe headers indicate the image is cymk while you know its rgb.

This code is a potential for crashes since we expect colorspace to match components and have such things in place.

Think of it like if you have a grayscale image, jpeg will only store one component, but were you to change input_colorspace to RGB, now the decoder expects there to be three components.

PartTimer1996:
Hi there, thanks for getting back to me.. So the adobe tag being set specifically suggests that the image is either CMYK or RGB, depending on the number of channels detected. In this case I know for a fact it is a three channel RGB image. From inspecting the metadata of the created jpeg it seems the suggest the tag is basically saying it's either RGB or CMYK and doesn't specify which (zune seems to detect this as CMYK) and this seems to be part of the Adobe jpeg spec (which is really annoying - link [here](https://docs.oracle.com/javase/6/docs/api/javax/imageio/metadata/doc-files/jpeg_metadata.html)). 

I know it's a case of the metadata being annoying and could prove error prone for users - but in the event the jpeg metadata could mislead the decoder like this, a manual override of detected properties at the hands of a user is useful when they know what the data they are parsing is made up of. 

Definitely understand it's not stable depending on user interaction so would it be okay if leave this fork of the code for my personal use case? Also, thanks for the package it's awesome and much quicker in production than the image decode alternative.

etemesi254:
Hi,  it may be a case of the library not handling adobe images correctly.

The code that does that is at 

https://github.com/etemesi254/zune-image/blob/bdd9caa0c6d0e288255afea4ada0467bbd4dbc1c/zune-jpeg/src/headers.rs#L441-L496

I wrongly assume that 0 is for CYMK but this mentions that transform 0 +3 color components is RGB (I never saw such literature for Adobe APP14)

So probably that may fix it, instead of introducing the new API.

It's not an easy problem to fix by just checking how many components we have because if Adobe App14 comes after the tag that tells us the number of components , we will have a wrong value

There is a function called `setup_component_params` in misc.rs called after reading headers but before decoding the image, 
would replacing it with the following fix your problem?

```Rust
pub(crate) fn setup_component_params(img: &mut JpegDecoder) -> Result<(), DecodeErrors>
{
    let img_width = img.width();
    let img_height = img.height();

    // in case of adobe app14 being present, zero may indicate
    // either CYMK if components are 4 or RGB if components are 3, 
    // see https://docs.oracle.com/javase/6/docs/api/javax/imageio/metadata/doc-files/jpeg_metadata.html
    // so since we may not know how many number of components
    // we have when decoding app14, we have to defer that check
    // until now.
    //
    // We know adobe app14 was present since it's the only one that can modify
    // input colorspace to be CYMK
    if img.components.len() == 3 && img.input_colorspace == ColorSpace::CMYK
    {
       
        img.input_colorspace = ColorSpace::RGB;
    }

    for component in &mut img.components
    {
        // compute interleaved image info
        // h_max contains the maximum horizontal component
        img.h_max = max(img.h_max, component.horizontal_sample);
        // v_max contains the maximum vertical component
        img.v_max = max(img.v_max, component.vertical_sample);
        img.mcu_width = img.h_max * 8;
        img.mcu_height = img.v_max * 8;
        // Number of MCU's per width
        img.mcu_x = (usize::from(img.info.width) + img.mcu_width - 1) / img.mcu_width;
        // Number of MCU's per height
        img.mcu_y = (usize::from(img.info.height) + img.mcu_height - 1) / img.mcu_height;

        if img.h_max != 1 || img.v_max != 1
        {
            // interleaved images have horizontal and vertical sampling factors
            // not equal to 1.
            img.is_interleaved = true;
        }
        // Extract quantization tables from the arrays into components
        let qt_table = *img.qt_tables[component.quantization_table_number as usize]
            .as_ref()
            .ok_or_else(|| {
                DecodeErrors::DqtError(format!(
                    "No quantization table for component {:?}",
                    component.component_id
                ))
            })?;

        let x = (usize::from(img_width) * component.horizontal_sample + img.h_max - 1) / img.h_max;
        let y = (usize::from(img_height) * component.horizontal_sample + img.h_max - 1) / img.v_max;
        component.x = x;
        component.w2 = img.mcu_x * component.horizontal_sample * 8;
        // probably not needed. :)
        component.y = y;
        component.quantization_table = qt_table;
        // initially stride contains its horizontal sub-sampling
        component.width_stride *= img.mcu_x * 8;
    }
    if img.is_interleaved
        && img.components[0].horizontal_sample == 1
        && img.components[0].vertical_sample == 1
    {
        return Err(DecodeErrors::FormatStatic(
            "Unsupported unsampled Y component with sampled Cb / Cr components"
        ));
    }

    // delete quantization tables, we'll extract them from the components when
    // needed
    img.qt_tables = [None, None, None, None];

    Ok(())
}
```

PartTimer1996:
Hi there, thanks for the reply. 

From reviewing the updated code you posted I think it would absolutely fix this issue I'm having. I'd be happy to give this a test on my fork unless you can want to create a feature branch? 

Thanks again.

etemesi254:
could you try to see if it works?

PartTimer1996:
Hi there, I tested out the change you made as suggested above on the dev branch and it works a treat for me! I'll close this PR and continue using the dev branch until it's merged into the main crate. Many thanks for this!

PartTimer1996:
Closing as suggested dev fix is working great.

PR #127, opened 2023-05-03
zune-hdr: Add support for no_std with Rust libm

JohnAZoidberg:
Based on #126

Currently float and exp2 are not available on no_std because std links
to the C library libm to get access to those. Currently there's a port
of libm to Rust underway, with the eventual goal of supporting float
math in core.

I haven't done benchmarks yet but I have seen suggestions (from a
couple years ago) saying it might be slower still than C libm.
Perhaps it might make sense to optionally depend on Rust libm only in
no_std environments.


etemesi254:
Hi,  I would prefer to have us reimplement floor and exp2 inside zune-hdr. Something I'm also careful about is dependencies. 

It also adds the advantage of making it faster since compiler can inline it.

For exp2, I wanted to use fastexp2 function 
http://www.machinedlearnings.com/2011/06/fast-approximate-logarithm-exponential.html

And the current implementation at https://github.com/etheory/fastapprox/blob/master/fastapprox/src/fastexp.h
but without the unions and unsafe, the accuracy is good enough since already hdr is a lossy codec.

And thanks for the PR

JohnAZoidberg:
Sure, i thought you might have reservations about this.

That's ok, unlike me other PRs, this one isn't something I'm actually using. So don't feel rushed to implement this.

etemesi254:
Should work on it over the weekend/after I finish a bmp decoder(soonish). 

Thanks

Issue #134, opened 2023-06-11
`zune-jpeg` Decoding to Luma colorspace gives bad results


Olwaro:
I'm trying to get a grayscale version of some pictures with the following snippet:

```rust
use zune_core::colorspace::ColorSpace;
use zune_core::options::DecoderOptions;
use zune_jpeg::JpegDecoder;

fn main() {
    let options = DecoderOptions::default().jpeg_set_out_colorspace(ColorSpace::Luma);

    let raw = std::fs::read("lena.jpg").unwrap();

    let mut decoder = JpegDecoder::new_with_options(options,&raw);

    let pixels = decoder.decode().unwrap();

    std::fs::write("grayscale.data", &pixels).unwrap();
}
```

I tried several pictures from different sources and I always get the same result. The decoded images are squeezed in the first half height-wise and the second half is all zeroes.

Exemple output with a random lena.jpg found online :

![grayscale](https://github.com/etemesi254/zune-image/assets/477023/90b67f29-633b-4ef8-9c6a-a3ad6dae841b)

Version:
```
├── zune-core v0.2.14
│   └── bitflags v2.3.1
└── zune-jpeg v0.3.17
    ├── log v0.4.19
    └── zune-core v0.2.14 (*)
```

etemesi254:
hi

thanks for the bug report


will look into it

etemesi254:
could you kindly attach the image from which the above output was gotten? JPEG format is complicated hence some images may work and others don't

Olwaro:
Sure, I should have done it from the start.
The original image was fetched from [here](https://bellard.org/bpg/lena30.jpg)

![lena30.jpg](https://github.com/etemesi254/zune-image/assets/477023/20a2175d-5b3d-4d87-8cf4-02729247a808)

etemesi254:
Hi, this is fixed in the dev branch commit a59c6753d7687dab0ef00389a7b54b7db8970d94

The issue happened to be that for sub-sampled images, `post_process` functions would work on two mcu widths (block of 8 by 8 pixels) but only increment one width, it seems I forgot to multiply by `h_max` which is a value that  (loosely) tells us how many mcu's we process by.

Olwaro:
Thanks for the fix and explainations !

PR #135, opened 2023-06-16
Handle MCU height other than 8 when decoding to Luma

Olwaro:
I noticed that the issue I openned (https://github.com/etemesi254/zune-image/issues/134) only happen when subsampling makes `mcu_height` larger than 8.

This is a quick fix.

### DIFF
--- a/zune-jpeg/src/mcu.rs
+++ b/zune-jpeg/src/mcu.rs
@@ -410,7 +410,7 @@ impl<T: ZReaderTrait> JpegDecoder<T>
             )?;
 
             // increment pointer to number of pixels written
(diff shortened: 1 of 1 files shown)


etemesi254:
Aaa sorry if it seems  I took long, usually use the weekends for such projects, thanks for the pull request.

Looking into it

Olwaro:
It's nothing like that. I just took the opportunity to try to understand jpeg decoding a bit more.

My solution is a bit simplistic, take your time to come up with something more polished 😊

etemesi254:
Would have worked for this case, but would probably break other grayscale images, but thanks for the pr nonetheless

PR #138, opened 2023-09-06
png: Make bytemuck optional

not-fl3:
For `zune-png`, `bytemuck`, the crate, is used just once, to convert u16 slice to u8 slice. 

This PR does not alter the default behavior: this conversion is still going through the crate. 
But without `bytemuck` feature it is just `slice.align_to_mut::<u8>()`. 

I tested the new code on the hdr PNG triggering this codepath and it all seems to still work fine.

New function looks ugly, but, unfortunately, attributes on regions are not yet stable and that was the simplest workaround I've seen.

_I do have a PR for the `log` in queue :D Not like `bytemuck` and `log` are that big, but having them under a feature-flag would be really nice!_

### DIFF
--- a/zune-png/src/decoder.rs
+++ b/zune-png/src/decoder.rs
@@ -27,7 +27,7 @@ use crate::filters::de_filter::{
 use crate::options::default_chunk_handler;
 use crate::utils::{
     add_alpha, convert_be_to_target_endian_u16, expand_bits_to_byte, expand_palette, expand_trns,
-    is_le
+    is_le, convert_u16_to_u8_slice
 };
 
 /// A palette entry.
@@ -777,11 +777,8 @@ impl<T: ZReaderTrait> PngDecoder<T> {
         {
             &mut out_u8
         } else {
-            let (a, b, c) = bytemuck::pod_align_to_mut::<u16, u8>(&mut out_u16);
+            let b = convert_u16_to_u8_slice(&mut out_u16);
 
-            // a and c should be empty since we do not expect slop bytes on either edge
-            assert!(a.is_empty());
-            assert!(c.is_empty());
             assert_eq!(b.len(), new_len * 2); // length should be twice that of u8
             b
         };
--- a/zune-png/src/utils.rs
+++ b/zune-png/src/utils.rs
@@ -350,3 +350,14 @@ pub(crate) fn add_alpha(input: &[u8], output: &mut [u8], colorspace: PngColor, d
         (a, b) => panic!("Unknown combination of depth {a:?} and color type for expand alpha {b:?}")
     }
 }
+
+pub fn convert_u16_to_u8_slice(slice: &mut [u16]) -> &mut [u8] {
+    // Converting a u16 slice to a u8 slice is always correct because
+    // the alignment of the target is smaller.
+    unsafe {
+        std::slice::from_raw_parts_mut(
+            slice.as_ptr() as *mut u8,
+            slice.len().checked_mul(2).unwrap(),
+        )
+    }
+}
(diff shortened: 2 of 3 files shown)


etemesi254:
Hi, thanks for your contribution.

I don't see any problem with the code, but pinging @Shnatsel  in case there is some memory bug introduced here.



I'm not a fan of nested functions so I'd rather the function be moved to be a standalone (probably in utils.h)

But in case there are no bugs, we should probably drop the bytemuck dependency all together

Shnatsel:
Looks good to me, incl. dropping bytemuck entirely. 

I'd like to see a more detailed safety comment. Calling out that alignment of u8 always less than u16, and both types are valid for all bit patterns with no lifetimes so a transmute is fine (`align_to` is [essentially a transmute](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.align_to_mut)).

Also we should leave a comment that if more such conversions are needed, `bytemuck` should be brought back. It's only dropped because this conversion is exceedingly trivial.

zune-png/src/decoder.rs
```diff
@@ -777,7 +777,22 @@ impl<T: ZReaderTrait> PngDecoder<T> {
         {
             &mut out_u8
         } else {
-            let (a, b, c) = bytemuck::pod_align_to_mut::<u16, u8>(&mut out_u16);
+            #[inline(always)]
```
Nilstrieb, line 780: This can be greatly simplified by just dropping all of this alignment logic. Converting a 16 slice to a u8 slice is always correct because the alignment of the target is smaller. Just casting it by hand and doubling the size is easier to understand than the align_to_mut dance imo.

Nilstrieb: The bytemuck https://docs.rs/bytemuck/latest/bytemuck/fn.bytes_of_mut.html function could be used but if bytemuck is optional it probably makes more sense to just drop it

zune-png/src/utils.rs
```diff
@@ -350,3 +350,9 @@ pub(crate) fn add_alpha(input: &[u8], output: &mut [u8], colorspace: PngColor, d
         (a, b) => panic!("Unknown combination of depth {a:?} and color type for expand alpha {b:?}")
     }
 }
+
+pub fn convert_u16_to_u8_slice(slice: &mut [u16]) -> &mut [u8] {
+    // Converting a u16 slice to a u8 slice is always correct because
+    // the alignment of the target is smaller.
+    unsafe { std::slice::from_raw_parts_mut(slice.as_ptr() as *mut u8, slice.len() * 2) }
```
Shnatsel, line 357: What if `slice.len() * 2` overflows?

Shnatsel: I don't think it can but there should be a comment explaining why it cannot

not-fl3: I also thought it can not, but fixed it with a `checked_mul`.

etemesi254:
> But this change brakes the promise of Safety: No unsafe code, with the sole exception of SIMD intrinsics which currently require unsafe. I thought that having bytemuck by default is a good compromise: still no unsafe in the zune codebase involved by default, but possible to avoid extra deps.

Fine by me, it's one well documented line

etemesi254:
> I do have a PR for the log in queue :D Not like bytemuck and log are that big, but having them under a feature-flag would be really nice!

For this, btw I'd love if it went to `zune-core`, maybe have a macro like `zune_warn!` which expands to log::warn! if a `log` feature is included in `zune-core` otherwise expands to nothing, then this can be used to disable logging for all libraries with one swoop,  just an opinion :).

etemesi254:
@not-fl3  Is this okay to merge?

not-fl3:
> @not-fl3 Is this okay to merge?

I believe so! Rebased out my first try commit.

etemesi254:
Thank you, will make a breaking release probably over the weekend with the changes

PR #141, opened 2023-09-14
Make log optional

not-fl3:
By default, with the "log" feature on "zune-core" will re-export the log crate.
Without "log", "zune-core" exports some stubs over `eprintln!`. 
The only change required for "zune-*" crate: use `zune_core::log::trace` instead of `log::trace`.

To remove "log-rs" dependency in zune-image user's Cargo.toml:
```
zune-jpeg = {version = "0.3", default-features = false, features = ["x86", "neon", "std"]}
```
I do not really like that it requires "x86" etc, but I guess its fine? Not having `log` as a default featulre will be a breaking change - for everyone who actually used `log-rs`, all the logs from `zune` will suddenly disappear, not ideal.

Let me know if the idea is OK, I'll finish this PR with the rest of "zune-*" crates.

### DIFF
--- a/zune-core/src/log.rs
+++ b/zune-core/src/log.rs
@@ -1,0 +1,68 @@
+#[repr(usize)]
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Hash)]
+pub enum Level {
+    Error = 1,
+    Warn,
+    Info,
+    Debug,
+    Trace,
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __log_enabled {
+    ($lvl:expr) => {{
+        let _ = $lvl;
+        false
+    }}
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __error {
+    ($($arg:tt)+) => (
+        #[cfg(feature = "std")]
+        {
+            panic!($($arg)+);
+        }
+    )
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __warn {
+    ($($arg:tt)+) => (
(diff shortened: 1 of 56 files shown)


etemesi254:
yea, it is ideal. I'm planning on a breaking change anyways since i switched from requiring slices for input to a generic trait T

not-fl3:
> yea, it is ideal. I'm planning on a breaking change anyways since i switched from requiring slices for input to a generic trait T

perfect! pushed the rest of the crates with the "log" being off by default. 

The only thing that concerned me - I changed log for zune_core in zune-imageprocs. Do not really know whats the relashionship between zune-imageprocs and zune_core and if that was a right thing to do.

etemesi254:
> The only thing that concerned me - I changed log for zune_core in zune-imageprocs. Do not really know whats the relashionship between zune-imageprocs and zune_core and if that was a right thing to do.

The reason it never used `zune-core` was that there was no need for them to interact.

The imageprocs does image processing stuff- blurs, color manipulation etc. core mainly provides decoding and encoding routines like byte readers and writers.

zune-core/src/log.rs
etemesi254, line 1: I'm not particularly okay with printing to standard output. I.e it's a library and it wouldn't make sense to print warnings for something we can't really influence where it would be used.

I think if the user doesn't want logging, let us just silently fail

not-fl3: done! warn and error are now panic and anything else is just nothing. 

tbh as a user I would not mind info! and warn! to go into stderr

PR #143, opened 2023-09-30
Remove bitflags dependecy

not-fl3:
I am not quite sure what to do with this change - making bitflags optional will not make much sense, while removing bitflags makes zune to consume more memory and does not make anything to work any better. 

However I figured that while I already made it work - better open a PR :)


etemesi254:
I think the size increase is warranted, reduces some complexity.

Maybe if we get really big we can switch back to bitflags

PR #170, opened 2024-03-20
Added functions to extend image with custom codecs

SalOne22:
Added functions `write_with_encoder` and `from_decoder`. This makes easier to integrate custom codecs without hacking the compiler like so
```rs
let mut decoder = AvifDecoder::try_new(reader)?;

return <AvifDecoder<ZByteReader<Vec<u8>>> as DecoderTrait<Vec<u8>>>::decode(
    &mut decoder,
);
```

### DIFF
--- a/crates/zune-image/src/codecs.rs
+++ b/crates/zune-image/src/codecs.rs
@@ -655,6 +655,30 @@ impl Image {
             ))
         }
     }
+
+    /// Open a new file from memory with the configured decoder
+    ///  
+    /// # Arguments
+    ///  - `decoder`: The configured decoder
+    /// # Example
+    /// - Open a memory source with the ppm decoder
+    ///
+    ///```no_run
+    /// // requires ppm feature
+    /// use zune_image::codecs::ppm::PPMDecoder;
+    /// use zune_image::image::Image;
+    ///
+    /// // create a simple ppm p5 grayscale format
+    /// let decoder = PPMDecoder::new(b"P5 1 1 255 1");
+    ///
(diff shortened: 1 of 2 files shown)


etemesi254:
Hi, thanks for the pr. :)

The decoder one is good, but the encoder one won't work with some changes I have in store,

specifically  I've recently been working on a new write api that allows one to write to write to abitrary streams, not just a `Vec<u8>` , i.e we can now write directly to anything that implements [`ZByteWriterTrait`](https://github.com/etemesi254/zune-image/blob/ae416a631590fd0b84c4ae44ffd8e49c0dd5fe3a/crates/zune-core/src/bytestream/traits.rs#L101) which includes the ability to write directly to files, and not an intermediate buffer and then a file

This means that the signature for writing becomes
```rust 
/// Write data to a sink using a custom encoder returning how many bytes were written if successful
/// or an error otherwise 
 pub fn write_with_encoder<T:ZByteWriterTrait>(
        &self, mut encoder: impl EncoderTrait,sink:T
    ) -> Result<usize, ImageErrors> {
        encoder.encode(self,sink)
}
```

instead of the implemented  one.

If implementations don't allow one to write to arbitrary buffers, one can write to a Vec or a slice and then copy it to the `sink` via `write_all_bytes` method.

etemesi254:
Sorry I think I added yours and mine by mistake. Thanks for that

etemesi254:
Sorry it seemed rebase couldn't work due to conflicts, had to squash and merge, hope that's okay (I'll fix the failing test in a few)

PR #200, opened 2024-05-11
fix miscalculated offset in Image::from_fn

hillin:
The offset calculation in `Image::from_fn` is incorrect.


etemesi254:
Thanks

PR #201, opened 2024-05-11
Better handling of EOF in read_exact_bytes

fintelia:
After e69b4c59a64292872790fa111bdeff140b690c4e I'm seeing that trying to read from a `BufReader` with a small buffer no longer works (using the wrapper from https://github.com/image-rs/image/pull/2198):

```rust
let decoder = JpegDecoder::new(BufReader::with_capacity(32, Cursor::new(include_bytes!("test.jpg")))).unwrap();
let mut decoded = vec![0; decoder.total_bytes() as usize];
decoder.read_image(&mut decoded).unwrap();
```

Produces:
```
I/O errors Not enough bytes, expected 14 but found 3144
```

The error is because `BufReader` sometimes returns fewer bytes than requested in a read call without it indicating that the end-of-input has been reached.

This PR changes the implementation of `ZByteReaderTrait::read_exact_bytes` to only treat zero bytes read as meaning EOF and all other partial reads get re-attempted

### DIFF
--- a/crates/zune-core/src/bytestream/reader/std_readers.rs
+++ b/crates/zune-core/src/bytestream/reader/std_readers.rs
@@ -15,23 +15,29 @@ impl<T: io::BufRead + io::Seek> ZByteReaderTrait for T {
     }
     #[inline(always)]
     fn read_exact_bytes(&mut self, buf: &mut [u8]) -> Result<(), ZByteIoError> {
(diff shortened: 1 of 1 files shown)


etemesi254:
Yea this makes more sense, thank you. 

Sorry for missing that, the docs are pretty weird on the behavior.

Issue #203, opened 2024-05-20
missing license texts for MIT and Apache-2.0 (and Zlib) from distributed crates


decathorpe:
I noticed that the crates that are published on crates.io from this project don't contain any license texts - upon further investigation, it looks like the project doesn't contain the MIT or Apache-2.0 license texts at all.

Both the MIT and Apache-2.0 licenses require that (re)distributed sources contain a copy of the license text, so it would be great if the MIT and Apache-2.0 license texts could actually be included in the repository, and all license files linked into crates so that cargo includes them when running `cargo publish`.

If a Pull Request to add the standard license texts for MIT and Apache-2.0 and adding symbolic links to include all license texts in published crates (as required) would be acceptable, I can work on that.

---

To provide some context: I am the maintainer of the packages for the `image` crate in Fedora Linux, and the update from image v0.24 to v0.25 added new dependencies on `zune-jpeg` and `zune-core`. I noticed that neither crate includes license texts for either MIT, Apache-2.0, nor Zlib.

The fact that the license texts are currently missing is currently blocking me from packaging them for Fedora Linux and updating the `image` crate (i.e. the Zlib license text is present in the repo but missing from published crates, and MIT and Apache-2.0 license texts are missing from the repo entirely).

etemesi254:
Hi , sorry missed that, they are in the top level repository, since the library is a group of crates

is it required to be in all crates?

If so a PR would be welcome

decathorpe:
> they are in the top level repository, since the library is a group of crates

Nope, license texts for MIT and Apache-2.0 are not present at all - https://github.com/etemesi254/zune-image/blob/dev/LICENSE.md is just a copyright statement but not a license text (it's also outdated since you added the Zlib option).

> is it required to be in all crates?

Yes, license texts are technically required to be present in all published crates. They *are* redistributed separately after all (via crates.io), so the fact that they originate in the same git repository is not really relevant.

> If so a PR would be welcome

Thanks, I will work on that later today.

etemesi254:
I believe this is now okay

PR #206, opened 2024-05-26
add license files for MIT/Apache-2.0 and include license texts in crates

decathorpe:
This PR is meant to address https://github.com/etemesi254/zune-image/issues/203. Sorry for the delay.

I have added the standard license texts for the MIT and Apache-2.0 licenses as `LICENSE-MIT` and `LICENSE-APACHE`, matching the existing `LICENSE-ZLIB` file.

The texts should be 100% matches for the standard license texts for the three MIT / Apache-2.0 / Zlib SPDX identifiers (see entries on choosealicense.com or opensource.org).

I have renamed the LICENSE.md file to COPYRIGHT.md, which seems to be the actual purpose of this file, and updated its contents to match the actual project license. The contents of the file still claimed the project was `Apache-2.0 OR MIT`, and did not mention the new `Zlib` option at all. This is now fixed.

I have verified that running `cargo package` for all crates in this workspace now results in the license texts being included in the files that are packaged by cargo, so they will also be included when `cargo publish` is run in the future.

Note that you might need to check that this works as expected if you do `cargo publish` from an OS that does not support symbolic links properly.

### DIFF
--- a/COPYRIGHT.md
+++ b/COPYRIGHT.md
@@ -1,0 +1,10 @@
+Copyright (c) zune-image developers
+
+Licensed under the Apache License, Version 2.0
+<LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0>, the MIT license
+<LICENSE-MIT or http://opensource.org/licenses/MIT>, or the Zlib license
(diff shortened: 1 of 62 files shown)


etemesi254:
Thanks

decathorpe:
Thank you!

PR #209, opened 2024-06-23
Fixed incorrect pixel position in Image::modify_pixels_mut

a16547:
Fix incorrect calculation of pixel position in Image::modify_pixels_mut.


etemesi254:
thank you

PR #213, opened 2024-07-10
Implement DecodeInto for more formats

Firestar99:
I just implemented the `DecodeInto` trait for more image formats that already have a `decode_into(buf: &mut [BufferType])` like function. For better image format support, I've also added `type BufferType` to `DecodeInto` so that eg. hdr can output in f32 instead of forced u8.

My use-case is to directly write into buffers allocated with vulkan that the GPU can directly read from, eliminating an extra copy from a `Vec<u8>`. Something `image-rs` does not at all support.

While making this I've noticed some minor inconsistencies between the crates:
* bmp: calls it `output_buf_size()` instead of `output_buffer_size()`, didn't change it as to not break compatability
* some decode errors seem to prepend the image format, some do not

### DIFF
--- a/crates/zune-image/src/codecs/png.rs
+++ b/crates/zune-image/src/codecs/png.rs
@@ -18,6 +18,7 @@ use zune_core::colorspace::ColorSpace;
 use zune_core::log::warn;
 use zune_core::options::EncoderOptions;
 use zune_core::result::DecodingResult;
+use zune_png::error::PngDecodeErrors;
 pub use zune_png::*;
 
 use crate::codecs::{create_options_for_encoder, ImageFormat};
@@ -27,7 +28,7 @@ use crate::errors::ImgEncodeErrors::ImageEncodeErrors;
 use crate::frame::Frame;
 use crate::image::Image;
 use crate::metadata::ImageMetadata;
-use crate::traits::{DecoderTrait, EncoderTrait};
+use crate::traits::{DecodeInto, DecoderTrait, EncoderTrait};
 
 impl<T> DecoderTrait for PngDecoder<T>
 where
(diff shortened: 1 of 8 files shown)


etemesi254:
Hi, appreciate this

> I just implemented the DecodeInto trait for more image formats that already have a decode_into(buf: &mut [BufferType]) like function. For better image format support, I've also added type BufferType to DecodeInto so that eg. hdr can output in f32 instead of forced u8.

Nice, I seemed to have forgotten about this trait power, hence why many weren't implemented since I didn't want to force them to  be decoded into `u8`

> bmp: calls it output_buf_size() instead of output_buffer_size(), didn't change it as to not break compatability
> some decode errors seem to prepend the image format, some do not

The naming can be changed, I'm doing a breaking `0.4`->`0.5` so they are free to be modified, the inconsitencies is because the crates are a lot and sometimes some inconsitencies arise.


> My use-case is to directly write into buffers allocated with vulkan that the GPU can directly read from, eliminating an extra copy from a Vec<u8>. Something image-rs does not at all support.

A valid usecase, nice
>

crates/zune-image/src/traits.rs
etemesi254, line 1: I think the documentation should now change to remove the uneeded information.
```text
 /// The rationale is that u8 bytes can alias any type
  /// and higher bytes offer ways to construct types from
   /// u8's hence they can be used as a base type
````
⬆️  should be removed

crates/zune-image/src/codecs/png.rs
etemesi254, line 1: Be wary of 16 bit images that may cause the buffer to be 2x the size expected and endianess may mess with color pixels. I'm not sure if that would be desired, should we explain in the documentation?.

Also we should add tests, but first if we can reach a consensus on the two concerns then we can add tests later

etemesi254: There are formats that I've implemented such support

1. JPEG and PNG have support to output to RGBA from RGB by adding an alpha channel with 255(opaque)

The problem with that is that it complicates output, for both formats it was code rarely called but included because someone needed it (mainly graphics programmers). Making it a requirement for all formats will complicate it.

An alternative to copy is to allocate the whole buffer capable of storing RGBA data, store RGB data there, and then pad and separate if necessary, e.g here's an example of an in-mut RGB to RGBA https://godbolt.org/z/rEoobYsxG (not confirmed to be working, also speed is not confirmed, but caches should make it fast)

Firestar99: Really only JPEG and PNG are needed as those are the two image formats most commonly used for game assets, eg. the glTF model format officially only supports these two image codecs. Them having support for RGBA output does indeed sound very interesting, and would be totally sufficient for my use-case!

etemesi254: Do note both have in built support for grayscale/1-component images, for such, activating the addition of an alpha component will convert it to Luma+Alpha

PR #239, opened 2024-10-16
[zune-jpegxl] Implement Display and Error for JxlEncodeErrors

Enet4:
I was interested in using this encoder directly instead of via `zune-image`, and tripped here while trying to handle the error that came. This PR adds impls of `core::fmt::Display` and `std::error::Error` (the latter gated on feature "std") for `JxlEncodeError` so that it can be used as an error directly.


etemesi254:
Thanks, I seemed to have forgotten that

Issue #251, opened 2025-02-21
Decoding fails with "Unknown marker 0xFF57"


tsxiaofang:
use anyhow::Result;
use std::{fs::read, io::Cursor};
use zune_jpeg::JpegDecoder;

fn main() -> Result<()> {
    // load some jpeg data
    let data = read("f:\\11.jpg").unwrap();
    // create a decoder
    let mut decoder = JpegDecoder::new(Cursor::new(data));
    // decode the file
    let _pixels = decoder.decode()?;

    Ok(())
}

Error: "Unknown marker 0xFF57"
error: process didn't exit successfully: `target\debug\img_example.exe` (exit code: 1)

sample image in 11.zip 
[11.zip](https://github.com/user-attachments/files/18901449/11.zip)

etemesi254:
Hi, what version are you on?

tsxiaofang:
I tested  zune-jpeg versions 0.4.14 and 0.5.0-rc2

Issue #246, opened 2024-12-30
Decoding fails with "Unknown marker 0xFF35"


sophie-h:
Upstream issue https://gitlab.gnome.org/GNOME/glycin/-/issues/105

Example file https://gitlab.gnome.org/-/project/27706/uploads/e3a2f9924387943ab87adf166224e2a6/P1000082.zip

o-tho:
I ran into the same issue (https://github.com/etemesi254/zune-image/issues/241). The solution was to use zune-jpeg 0.5 instead of the stable release.

etemesi254:
Should be fixed in 0.4.15 and 0.5.0-rc2

PR #252, opened 2025-02-26
zune-jpeg skip unknown marker instead of errors.

tsxiaofang:
A better choice would be to skip unknown markers
Related issues:
[https://github.com/etemesi254/zune-image/issues/251](url)
[https://github.com/etemesi254/zune-image/issues/246](url)


etemesi254:
A better way to handle that is to  show error in case one wants strict conformance other than ignore since the presence of an unknown marker in the bistream is an error in itself

etemesi254:
But lgtm for now

PR #253, opened 2025-02-27
AVX2, scalar YUV -> RGB improvements

awxkee:
Please, note comments on lane 261 in avx.rs.

This will be a little slower, almost no doubts. However, it has almost 3 times better precision, exactly as in libjpeg-turbo.

### DIFF
--- a/crates/zune-jpeg/src/color_convert.rs
+++ b/crates/zune-jpeg/src/color_convert.rs
@@ -57,7 +57,9 @@ pub use crate::color_convert::avx::{ycbcr_to_rgb_avx2, ycbcr_to_rgba_avx2};
 use crate::decoder::ColorConvert16Ptr;
 
 mod avx;
+mod neon64;
 mod scalar;
+
 #[allow(unused_variables)]
 pub fn choose_ycbcr_to_rgb_convert_func(
     type_need: ColorSpace, options: &DecoderOptions
@@ -74,6 +76,17 @@ pub fn choose_ycbcr_to_rgb_convert_func(
             match type_need {
                 ColorSpace::RGB => return Some(ycbcr_to_rgb_avx2),
                 ColorSpace::RGBA => return Some(ycbcr_to_rgba_avx2),
+                _ => () // fall through to scalar, which has more types
+            };
+        }
+    }
+    #[cfg(all(feature = "neon", target_arch = "aarch64"))]
+    {
+        if options.use_neon() {
+            use crate::color_convert::neon64::{ycbcr_to_rgb_neon, ycbcr_to_rgba_neon};
+            match type_need {
+                ColorSpace::RGB => return Some(ycbcr_to_rgb_neon),
+                ColorSpace::RGBA => return Some(ycbcr_to_rgba_neon),
                 _ => () // fall through to scalar, which has more types
             };
         }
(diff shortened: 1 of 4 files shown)


etemesi254:
Hi, thanks for this.

Did you manage to do any benchmarks of before after?

etemesi254:
> Please, note comments on lane 261 in avx.rs.

To get decode test for rgba you can use the method in options [jpeg_set_out_colorspace](https://docs.rs/zune-core/0.5.0-rc2/zune_core/options/struct.DecoderOptions.html#method.jpeg_set_out_colorspace) and start the decoder with the new options

```rust
// not written to work, just demonstrative 
use zune_core::options::DecoderOptions;

let options = DecoderOptions::new_fast().jpeg_set_out_colorspace(ColorSpace::RGBA)

let decoder = JpegDecoder::new_with_options([SOURCE],options)
```

awxkee:
> Hi, thanks for this.
> 
> Did you manage to do any benchmarks of before after?

No. I'll try to do benchmarks later.
I wouldn't expect this to be faster. I'd expect here at least 10-15% slowdown on YUV->RGB/RGBA path.

Not sure if workflow failures are related to this PR.

etemesi254:
> Not sure if workflow failures are related to this PR.

Hi, they are, the jpeg decoder tests were hash based so change in algorithms meas the hash changed, we can update that after merging so no major issues there

awxkee:
Also I'd like to note that coefficients computed by me is slightly different from the one [libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/adbb328159b5558e846690c49f9458deccbb0f43/simd/arm/aarch64/jsimd_neon.S#L103) does use.
At start, I preffered compute them by myself with MPFR 150 bits precision, because coefficients might be a legal issue, as we don't have the way to reproduce it. However, if you feel that we're safe and prefer exact match we could just use libjpeg-turbo coefficients.
However, libjpeg-turbo implementations also do not consistent and ARM coefficients and rounding modes do not matches that [x86/x86-64 is use](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/adbb328159b5558e846690c49f9458deccbb0f43/simd/i386/jdcolext-avx2.asm#L122).

awxkee:
Benchmarks:

Scalar implementation definetely lowers 5-12% of performance.
x86_64 AVX2 -2% - +2% that is almost noise level.
I've added NEON implementation to avoid slowdown the second major platform. It also does some slowdown, however on m3 laptop it is hard to say how many exactly, somewhere up to 5%.
Here is doc https://docs.google.com/document/d/1iFEsKD6-E9PUHik36MxitzoxWeOGphZAEZ7EXogvuQw/edit?usp=sharing

etemesi254:
Goood enough for me

awxkee:
If you want, I can do something as in libjpeg-turbo x86, it’ll almost half instruction count, and definitely will improve speed up to 15-20%. However, there will be trade off that results will be inconsistent between platforms, and scalar implementation. Seems like this works this way in libjpeg-turbo. This may work only for SIMD, for scalar implementation i32 arithmetic is required so there is nothing to do, without obvious precision lowering.

If you prefer this implementation, then probably it’s done.

I’ll try to bench aarch64 on stable cpu later, I think actually there is also should be -+2%

etemesi254:
I put accuracy above performance

and also C lI would fail due to inconsistent outputs so this is better for me

awxkee:
I tested out on c6g.large AWS Graviton. My initial guess about loss on aarch64 also was correct, it shows loss about ~1.5%.

PR #255, opened 2025-03-01
Gainmap metadata

awxkee:
You better check this with care, I'm not the very best in JPEGs parsing


awxkee:
Perhaps we could change ` pub gain_map_info: Vec<GainMapInfo>,` to ` pub gain_map_info: Option<GainMapInfo>,` if there is no plans to parse the whole image at one time ?
It doesn't look like is expected on libultrahdr to have multiple gain map metadata stored in the one image.

awxkee:
[Here](https://github.com/awxkee/gainforge/blob/01067c47eb95703952b262939c2f43aa8938ffec/src/iso_gain_map.rs#L276) is MPF parser, looks like everything works.

Issue #249, opened 2025-01-30
JPEG decoding yields discolored image (basic 8-bit YUV 444 JPEG with no / GIMP default sRGB color profile).


gergo-salyi:
I'm relaying this from my issue at the `image` crate using zune-jpg for JPEG decoding: https://github.com/image-rs/image/issues/2411

Tested on Linux with x86_64 CPU (i5-1035G4) 10th gen Intel having AVX2, etc...

```bash
# Convert with ImageMagick for reference
magick start.jpg good.png
```

```toml
# Cargo.toml
[dependencies]
image = "0.25.5"
```

```rust
// main.rs
fn main() {
    image::open("start.jpg").unwrap().save("bad.png").unwrap();
}
```

`start.jpg`: 

![Image](https://github.com/user-attachments/assets/83932746-a44c-443a-8a7d-2f7d6955551e)

`good.png`: 

![Image](https://github.com/user-attachments/assets/5275d535-03e3-4bbe-b1a6-ee312f1587f8)

`bad.png`: 

![Image](https://github.com/user-attachments/assets/12a6084c-8422-46a9-81f0-bcdf9e7ddd9f)

Lower half of `bad.png` is visibly more green then `start.jpg` or `good.png`. E.g. pixel at index `(150, 200)` became `#131910` instead of `#151811` . I expect the the JPEG decoder to yield identical result to ImageMagick.

Not sure about the cause, but I suspect a YCbCr -> RGB conversion problem. `zune-jpeg` does this: 

```rust
//! 1. The YCbCr to RGB use integer approximations and not the floating point equivalent.
//! That means we may be +- 2 of pixels generated by libjpeg-turbo jpeg decoding
//! (also libjpeg uses routines like `Y  =  0.29900 * R + 0.33700 * G + 0.11400 * B + 0.25000 * G`)
```

https://github.com/etemesi254/zune-image/blob/c9f333dd3f725e5fd044e0e6af37f2807485d35e/crates/zune-jpeg/src/color_convert/avx.rs#L16-L18

This integer approximation is known to have caused problems elsewhere in the past, see: https://en.wikipedia.org/wiki/YCbCr#Approximate_8-bit_matrices_for_BT.601

The [JPEG File Interchange Format Version 1.02 spec](https://www.w3.org/Graphics/JPEG/jfif3.pdf) defines on page 3:

```
RGB can be computed directly from YCbCr (256 levels) as follows:
R = Y + 1.402 (Cr-128)
[...]

PR #260, opened 2025-05-03
crates/jpeg/idct: Fix bad coefficient rounding on the no AC coeff fast paths

gergo-salyi:
IDCT calculations (all scalar and SIMD) on the fast paths with no AC coefficients incorrectly scaled the DC coefficient without rounding

This patch applies the correct formula with proper rounding

Fixes the remaining parts of issue:
https://github.com/etemesi254/zune-image/issues/249


kcking (approved):
Nice find! This vastly improved PSNR on this [test image](https://www.nasa.gov/wp-content/uploads/2025/04/jsc2025e034746orig.jpg) compared to turbojpeg from `0.00068149 (1.41492e-05)` to `4.20471e-06 (8.72986e-08)`! Would love to get this released and propagated to the `image` crate :D

PR #264, opened 2025-05-19
`no_std` Tidy-up

bushrat011899:
# Objective

- Split off low-controversy component of #263

## Solution

__Refer to individual commits for further details__

- Added `rust-version = "1.81.0"` to all crates, making it easier to maintain MSRV going forward. Exact version chosen for compatibility with `core::error::Error`.
- Switched to `core` implicit prelude instead of conditionally changing the prelude based on a possible `std` feature. This makes migrating currently `std`-only features to `no_std` easier in the future.
- Updated CI to test all `no_std` compatible features, not just `--no-default-features`.
- Updated `zune-core` to allow `serde` feature in `no_std`
- Updated `zune-inflate` to remove `std` feature (unused)
- Updated `zune-jpeg` to remove `std` feature (unused)
- Updated `zune-qoi` to remove `std` feature (unused)
- Updated `zune-bmp` to remove `std` feature (unused), opting for adding `std` feature in `dev-dependencies` instead
- Made `zune-gif` `no_std` (only minor changes required)
- Updated `zune-jpegxl` to explicitly require `std` in the `threads` feature

### DIFF
--- a/crates/zune-inflate/Cargo.toml
+++ b/crates/zune-inflate/Cargo.toml
@@ -15,10 +15,8 @@ rust-version = "1.81.0"
 [features]
 zlib = ["simd-adler32"]
 gzip = []
-std = ["simd-adler32/std"]
 
-
-default = ["zlib", "gzip", "std"]
+default = ["zlib", "gzip"]
 
 [dependencies]
 simd-adler32 = { version = "0.3.4", optional = true, default-features = false }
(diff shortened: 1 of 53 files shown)


crates/zune-inflate/Cargo.toml
etemesi254, line 1: This makes us lose runtime detection so may be slower see docs on  https://docs.rs/simd-adler32/latest/simd_adler32/

Any reason why it was removed

etemesi254:
ping on comment above

PR #272, opened 2025-06-15
Properly sets marker after stream end

### LINKED ISSUE #270: zune-jpeg v0.4.17 fails to decode some images: "No more bytes"
These two images fail to decode on zune-jpeg v0.4.17, but work fine in `imagemagick` and Firefox: [zune-jpeg-4.17-decode-failures.zip](https://github.com/user-attachments/files/20656713/zune-jpeg-4.17-decode-failures.zip)

v0.4.14 did not report an error, but did not decode them correctly either - the resulting output was very pixelated.

awxkee:
After stream end marker of stream end wasn't set so it causes excessive reads.
Run rustfmt also on that file.

This is not complete fix, it only returns the behaviour of previous versions. Therefore, it does only partial progressive decoding. 

Closes #270

### DIFF
--- a/crates/zune-jpeg/src/mcu_prog.rs
+++ b/crates/zune-jpeg/src/mcu_prog.rs
@@ -32,7 +32,7 @@ use crate::bitstream::BitStream;
 use crate::components::{ComponentID, SampleRatios};
 use crate::decoder::{JpegDecoder, MAX_COMPONENTS};
 use crate::errors::DecodeErrors;
-use crate::headers::{parse_sos};
+use crate::headers::parse_sos;
 use crate::marker::Marker;
 use crate::mcu::DCT_BLOCK;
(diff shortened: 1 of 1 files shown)


awxkee:
I'm not sure about CI fails, I don't think this patch should create a release.

etemesi254:
which scenarios does this change?

I am not sure that is the right fix for the issue tagged


the issue #270 is about incorrect handling of restart markers in the bitstream of Progressive images,  solved most of it over the weekend but still the image is not being decoded correctly. 

Would prefer to get it to decode correctly then look into what this PR does hopefully before end of this week

awxkee:
It fixes an attempt to read a buffer after it returns an error from `get_marker` if strict_mode is not enabled.

No it doesn't fixes completely issue only returns it to previous state where it decodes an image but it gets pixellated, because as you mentioned markers are not properly restarted.

PR #279, opened 2025-06-29
Fix mod path for compiling neon64.rs with rustc 1.84

r41k0u:
While running cargo tests for zune-jpeg with neon feature enabled (with Rust 1.84), we run into:  error[E0433]: failed to resolve: use of undeclared crate or module `std` The import should be from core::arch instead of std::arch

Test log: https://objectstorage.prodstack5.canonical.com/swift/v1/AUTH_0f9aae918d5b4744bf7b827671c86842/autopkgtest-questing/questing/arm64/r/rust-zune-jpeg/20250628_130958_25317@/log.gz


etemesi254:
Sorry, missed that.

Thanks for the pr

PR #284, opened 2025-07-27
Prefer the use of div_ceil, error messages on components

197g:
Using `div_ceil` (rust-version 1.73) avoids any concerns of overflows that might occur while running this operation and is, imo, far easier to read by avoiding the duplication of divisor. While reviewing these I came across two error messages related to components that were incongruent to the code they error'd on.

### DIFF
--- a/crates/zune-jpeg/src/headers.rs
+++ b/crates/zune-jpeg/src/headers.rs
@@ -281,14 +281,14 @@ pub(crate) fn parse_start_of_frame<T: ZByteReaderTrait>(
 
 /// Parse a start of scan data
 pub(crate) fn parse_sos<T: ZByteReaderTrait>(
-    image: &mut JpegDecoder<T>
+    image: &mut JpegDecoder<T>,
 ) -> Result<(), DecodeErrors> {
     // Scan header length
     let ls = image.stream.get_u16_be_err()?;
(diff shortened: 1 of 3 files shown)


197g:
I'm not sure I understand why CI was running (and failing) release checks here?

```
on:
  pull_request:

```

Does that match all PRs?

etemesi254:
The CI fail i think is because i wanted cargo release to generate binaries for me using https://github.com/axodotdev/cargo-dist, but never got to set it up properly.

Thanks for the PR, i think when I started the function was not present, not sure what my msrv is but going back 3 years has seems overkill

etemesi254:
On CI, pending items


 - Compile and test on more archs trying to acquire bit identical execution (ensure things like SIMD paths produce same values), was thinking of a way of achieving this just for neon now, but not yet  sure if its possible.
  -  Also get the `cargo dist` to work, i think removing the `on-pull` tag.
  -  Compile for WASM (a better check for no-std support )
  - Add msrv test ( last 5 versions)
  - Add python checks, (just building now), there is a file `./workflows/zune-python.yml` that does not run on CI
  
On documentation on benchmarks

- Ensure you are in the root directory, not a sub-directory otherwise it will not run, but I see where the confusion can occur so an update is welcome

Issue #277, opened 2025-06-21
JPEG: images decode broken for some


xmakro:
At `dev` branch head. Some decoding issues:

Input JPEG:
![Image](https://github.com/user-attachments/assets/17b28bf4-1135-4007-a2bf-3ac31103990e)

Output PNG:
![Image](https://github.com/user-attachments/assets/239bb65b-3880-4b5b-9a1c-cb8de7e5f105)

etemesi254:
Yikes, thats pretty bad.

Looking into it

197g:
The problem here first appears as early termination, in the 49'th MCU row (which I believe fits with where the black region starts):

```
ERROR [zune_jpeg::mcu] Marker `SOS` Found within Huffman Stream, possibly corrupt jpeg
TRACE [zune_jpeg::headers] Ss=0, Se=63 Ah=0 Al=0
WARN  [zune_jpeg::mcu] Got terminate signal, will not process further
```

However, I don't think that's the problem. Indeed patching this over just makes things worse and completely breaks the whole thing. Also note that colors are already way off and this is _not_ a progressive image, so all previous blocks are broken. Instead, I think the fault lies in non-progressive mode ignoring `z_order`. Since this has an effect on huffman table choices it will mess with the amount of bitstream data being consumed, which is not self-synchronizing. Using the component iteration order from `z_order` instead results in this:

<img width="1199" height="799" alt="Image" src="https://github.com/user-attachments/assets/7cc57d57-9999-4125-b440-bd4d97a0e0be" />

Still wrong, but the outline is there. I believe it's just a mixup of color components based on my rudimentary fix of the fault.

197g:
It matches up at my end:


```diff
+        let z_order = self.z_order;
+
         for j in 0..mcu_width {
             // iterate over components
-            for component in &mut self.components {
+            for &k in &z_order[..usize::from(self.num_scans)] {
+                let component = &mut self.components[k];
```

Then we get the correct scan. However, since the image is _also_ subsampled and the loop for baseline always uses the `mcu_x` instead of considering the mcu_s based on the present scans it then runs off the end of the second component.

```text
JFIF APP0 marker: version 1.01, density 1x1  0
Start Of Frame 0xc0: width=1199, height=799, components=3
    Component 1: 2hx1v q=0
    Component 2: 1hx1v q=1
    Component 3: 1hx1v q=1
Define Huffman Table 0x00
Define Huffman Table 0x10
Define Huffman Table 0x01
[...]

PR #286, opened 2025-07-28
Support baseline with multiple SOS passes

197g:
Closes: #277 

Please benchmark and review closely, there's subtle details here. I've added a few trace calls of my own that helped me in narrowing down the source of the issue. My own benchmark results are almost neutral.

### DIFF
--- a/crates/zune-jpeg/src/mcu.rs
+++ b/crates/zune-jpeg/src/mcu.rs
@@ -166,49 +166,108 @@ impl<T: ZByteReaderTrait> JpegDecoder<T> {
             }
         }
 
+        // If all components are contained in the first scan of MCUs, then we can process into
+        // (upsampled) pixels immediately after each MCU, for convenience we use each row of MCUS.
+        // Otherwise, we must first wait until following SOS provide the remaining components.
+        let all_components_in_first_scan = usize::from(self.num_scans) == self.components.len();
+        let mut progressive_mcus: [Vec<i16>; 4] = core::array::from_fn(|_| vec![]);
+
+        if !all_components_in_first_scan {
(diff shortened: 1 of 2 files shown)


crates/zune-jpeg/src/mcu.rs
```diff
@@ -284,57 +430,71 @@ impl<T: ZByteReaderTrait> JpegDecoder<T> {
...
-                    }
-                    error!(
-                        "Marker `{:?}` Found within Huffman Stream, possibly corrupt jpeg",
-                        m
-                    );
-                    self.parse_marker_inner(m)?;
-                    // if m == Marker::SOS {
-                    //     return Ok(true);
-                    // }
+            if stream.marker.is_some() && stream.bits_left == 0 {
```
etemesi254, line 441: not sure about this in regards to us pre-filling the byte buffer (as explained in https://github.com/etemesi254/zune-image/issues/277#issuecomment-3131350182) but lets see if testing brings up any anomalies

etemesi254 (approved):

etemesi254:
> Re tests, I think they broke in https://github.com/etemesi254/zune-image/commit/748dbb1909492f070734167df7eb66ec09ae5921 and everything else is unrelated to the PR.

I already reverted that error, it was from an incorrect assumption that the reason the image this PR fixes was failing to decode was an invalid SOS length, but it was because we never reset the marker so the decoder treated anything after that marker data.

etemesi254:
> so with the current code any errors you might get in development will be hard to interpret before that is fully checked (since I stumbled with that function, too, more unit test coverage of it would make that a more pleasant experience to develop).

Do you suggest unit tests of the marker reading functions?

PR #287, opened 2025-08-29
zune_jpeg: Add capabilities to access the IPTC metadata

1c3t3a:
IPTC metadata is stored in the APP13 blog and starts with the prefix "Photoshop 3.0". This change adds capabilities for zune_jpeg to parse this data and expose it in the public Info API.

I tested this patch locally and was able to successfully extract the image metadata. However, I didn't understand this crate's test-suite enough to add a test here and saw that simillar tests for exif were removed. I am open to add a test for this patch and would appreciate some guidance on how to best do that! :)


etemesi254:
Thank you

etemesi254:
Done, 0.4.21 for image

1c3t3a:
Hey @etemesi254, I noticed that the release somehow doesn't contain the latest version of the code and notably not the XMP and IPTC methods on the decoder. It's interesting because the repository link points to the correct version, but the actual packaged version doesn't have the methods. https://docs.rs/zune-jpeg/0.4.21/zune_jpeg/struct.JpegDecoder.html for reference.

1c3t3a:
Also a quick question regarding the other relase candidates: Would you recommend to use them? the latest one seems to have IPTC metadata, does it make any sense for the image crate to depend on this?

etemesi254:
it is in the info struct

https://docs.rs/zune-jpeg/0.4.21/zune_jpeg/struct.ImageInfo.html

etemesi254:
> Also a quick question regarding the other relase candidates: Would you recommend to use them? the latest one seems to have IPTC metadata, does it make any sense for the image crate to depend on this?

The release candidates are stable

the major difference is in I/O where the model changed from buffering up all input into a slice to normal 'Read' rust type (so you can start decoding even without full image data)

PR #299, opened 2025-10-26
Specialized IDCT for 4x4 filled blocks

197g:
This adds a specialized inverse cosine transform that detects and relies on several coefficients being `0`. Many of the operations reduce in several ways when the input matrix is only sparsely filled in the top. 

This mainly affects compressed images with banded coefficients, whose coefficients are sparse on purpose. The fraction of MCUs that are sparse in this way can be ~20%.

## Approach

When considering the IDCT equations, it is a sped up matrix multiplication and essentially linear. The first half of the algorithms keeps zeroed columns (note: we immediately transpose, so rows in the code) zeroed. Also several terms drop out if we prove they multiply with zeroed coefficients. We insert zero constants where they appear and propagate terms through the 8x8 implementation accordingly, one term at a time and one identity transform at a time. The final algorithm is not easily recognized due to several factors:

- many intermediate values are simple renaming due to adding zero
- some pairs of intermediates values can be identified with each other, the cosine transform does a bunch of terms looking like `(c1, c2) = (a + b, a - b)`. If `b` is zero then `c1 = c2`.
- some chains of additions are easily collapsed, i.e. some magic constant coefficients combine especially after renaming. In the full 8x8 the aim of reducing multiplications motivates using mutable intermediates but that seemed no longer the case. I've tried to never increase operation counts during individual refactors. Further, of course we keep potential overflows in mind, not all addition reorders are trivial.

Work-in-Progress: I've not fully cleared up the algorithm yet. I'd also want to apply more thorough SIMD considerations to the AVX version.

## Use

Work-in-Progress.

There already is one such case for `1x1`, detected by loading all indices in the 8x8 IDCT code and checking for 63 zeros. Checking the 4x4 case could be done similarly but that would influence the code path of all blocks. Instead, it can also be detected during MCU bitstream decoding. Recall that this in its core performs a sequence of skip-append operations of coefficients. We naturally get the last index that has been inserted. Since the indices from this operation map directly into matrix cells (via zig-zag encoding), a simple index bound of `10` (i.e. the first 4 diagonals) catches many of the 4x4 sparse cases with little effect on the other path.

```
|0|2|3|9|…|
|1|4|8|…| |
|5|7|…| | |
|6|…| | | |
|…| | | | |
```

Comparing both approach a little more in-depth would be favorable however.

## Benchmarks

Similar to the 8x8 case, there's a scalar version on which the AVX vectorization is based directly on. Benchmarks for now show ~2-4% improvements end-to-end, some outliers of -1% in vertical subsampling. That may just be unreliable, it show improvement in the mozjpeg comparison... Benchmarks are also pending the above other approaches to detect and dispatch into the specialized idct. The no-intrinsics run should be highlighted with +9.5% improvements, which should be very noticable on non-x86 targets.


etemesi254:
beautiful ✨✨✨

197g:
I'm going to have to look at the hash mismatches in the x86 build. The function should be bit-for-bit identical afterall.

197g:
I intentionally wanted to avoid adding it as a parameter. That pushes the conditional check _downwards_ which is arguably worse for optimization. We could also look into doing a bulk-idct that runs multiple blocks internally instead of multiple calls. This may have potential benefits:

- with SIMD register use (some XMMs are for arguments in the ABI)
- keep constants loaded
- maybe allow using non-temporal scatter stores in the outputs (needs an mfence at the return that would be amortized)
- some more opportunity for instruction interleaving, we have pipeline stalls at a few locations right now and the function is not memory bound.

For bulk operations however we definitely want to first group blocks by their length so that we have separate bulk-8x8, bulk-4x4, bulk-1x1 instead. Doing another internal dispatch in these functions would mostly make interleaving more difficult.

etemesi254:
|benchmark|before|after|
|-----------|------|-----|
|No sampling Baseline decode/zune-jpeg|54.536 ms| 50.798 ms|
|Horizontal Sub Sampling/zune-jpeg|47.757 ms| 47.101 ms |
|Vertical sub sampling/zune-jpeg|46.751 ms| 46.149 ms|
|HV sampling/zune-jpeg|43.528 ms|43.595 ms|

Machine Macbook M3.

Got by running `cargo bench --workspace "jpeg"`

Some nice improvements in [1], the benefits may not be that high on the set of images since they are high quality and hence few blocks will be having few DCT items.

But overall a worthwile optimization, appreciate you taking your time to do it.

In case of anything you can point out here or raise an issue.

PR #312, opened 2025-12-03
Fix handling of truncated images

mhils:
This PR improves how zune-jpeg handles truncated files. More specifically:

1. Fix premature end detection: While 37 is a wonderful lucky prime, I think the previous detection logic didn't quite work. At least I could never get that branch to trigger, and I had some images for which it (incorrectly) never triggered.
My amateur understanding is that `overread_by` is only ever increased by `usize::from(reader.eof()?);`, so I don't quite understand how we could produce false negatives here. Testing seems to confirm that. Please take a look, you have more expertise here. :)
2. Fill the remaining image with gray pixels. This is often visually nicer (and also matches what libjpeg-turbo does). Happy to take that part out if you think it doesn't make sense.

Thanks!

### DIFF
--- a/crates/zune-jpeg/src/mcu.rs
+++ b/crates/zune-jpeg/src/mcu.rs
@@ -199,12 +199,8 @@ impl<T: ZByteReaderTrait> JpegDecoder<T> {
             trace!("Decoding MCU width: {mcu_width}, height: {mcu_height}");
 
             for i in 0..mcu_height {
-                // Report if we have no more bytes
-                // This may generate false negatives since we over-read bytes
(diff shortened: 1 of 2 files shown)


etemesi254:
Hi, could you share before and after images of "incomplete_image" ,

you can do that with `cargo run -- -i {INCOMPLETE_PATH}  -o "incomplete_img_before.jpg"` (on the root workspace),  currently away from my machine but would not want to keep this waiting if it fixes something

etemesi254:
Nice, the latter is way better.

Thank you for this

PR #317, opened 2025-12-09
AVX2 Upsampling

mhils:
This PR adds optimized AVX2 upsampling. Decode time goes down 50%-70% on my benchmarks. 🥳

### DIFF
--- a/crates/zune-jpeg/Cargo.toml
+++ b/crates/zune-jpeg/Cargo.toml
@@ -21,6 +21,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(fuzzing)'] }
 x86 = []
 neon = []
 std = ["zune-core/std"]
+portable_simd = []
 log = ["zune-core/log"]
 default = ["x86", "neon", "std"]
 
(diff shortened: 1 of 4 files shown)


etemesi254:
Hi, awesome work 🥳🥳

Could you run the following and paste we see benchmarks before and after on your machine

```shell
 cargo bench --workspace "jpeg"
 ```
 
 I sadly don't have an x86 machine with me, so can't test.
 
 Then we can look into the code

**Comment by mhils** [comment]
```
jpeg: HV sampling/zune-jpeg
                        time:   [57.921 ms 58.322 ms 58.734 ms]
                        thrpt:  [16.638 MiB/s 16.755 MiB/s 16.871 MiB/s]
                 change:
                        time:   [-15.203% -14.087% -13.011%] (p = 0.00 < 0.05)
                        thrpt:  [+14.957% +16.397% +17.929%]
```


I've added a portable_simd implementation that works almost exactly the same, that one should allow you to test. 😄

etemesi254:
Nice

on it


But thinking, is there a way we can do both vertical and horizontal sampling in one pass or is that too complicated?

etemesi254:
Hi, sorry

Merged this by mistake, was not familiar with working with PRS in the IDE.

Do note that rustc autovectorizes very well some of the code, seee

https://godbolt.org/z/1z87Efe7r vis a vis handwritten simd https://godbolt.org/z/d3GcPn9Ko 

Also the gains arent translated on my machine.

Will get an x86 machine and test, if none, will have to revert this

PR #319, opened 2025-12-09
Fix CI & Nits

mhils:
Following up on https://github.com/etemesi254/zune-image/discussions/316, this PR brings fixes some low-hanging fruits. :)


etemesi254:
Hashes should never diverge, means something is wrong/diverges in decode paths

etemesi254:
CI is back to green

PR #342, opened 2026-01-17
jpeg: Fix non-interleaved decoding with subsampling

### LINKED ISSUE #341: Incorrect output for non-interleaved JPEGs with subsampling
Non-interleaved baseline JPEGs with subsampling decode incorrectly when ceil(width / (8 * h_max)) * h_sample ≠ ceil(width / 8)

Example 1000x1500 image:

Original:

Re-encoded:

barronpm:
Fixes #341 

I have also added a new test case. Let me know if you need anything else


etemesi254:
thanks for this

looking into it

is that the only image that you know triggers this?

barronpm:
I originally discovered this on the image in #341, then made a another test image which I also reproduced it with. It should happen for any baseline, non-interleaved jpeg with the right dimensions and subsampling. (For 4:2:2 like these images if width % 16 is in [1, 8])


Before vs After of change on the test image:
test-broken
test-fixed

Issue #273, opened 2025-06-15
Some JPEGs are decoded incorrectly


Shnatsel:
Affected images with the highest divergence from the reference: [scraped_jpeg_zune_0.4.17_divergences.zip](https://github.com/user-attachments/files/20748010/scraped_jpeg_zune_0.4.17_divergences.zip)

I am seeing two main patterns of artifacts:

![Image](https://github.com/user-attachments/assets/eb5ddb69-ce0b-4e3b-a22a-f6dd20b9eefa)

![Image](https://github.com/user-attachments/assets/ec8f6037-30fe-4212-a4f6-93a19a2baac0)

and this grid pattern

![Image](https://github.com/user-attachments/assets/d6eb077a-d64e-4d9f-a94b-1ac11644ead0)

![Image](https://github.com/user-attachments/assets/f12e3f68-5096-4f5b-9579-b238a6079e07)

Tested on zune-jpeg v0.4.17

etemesi254:
Thanks, for the first I see its a subsampling issue, so that will be looked into.

For the second both zune-jpeg and libjpeg-turbo (via command `djpeg  -verbose -outfile ./problem.bmp "/Users/etemesi/Downloads/errors2.jpg"` ) report premature image end. So provably we handle it a bit different, will look at where divergences occur and fix.

But may take a while (end of week) as I am working on the previous issue (exhausted data) so when that is solved will look into this

Shnatsel:
As of v0.5.5 the first image is fixed, while the second one still shows artifacts.

PR #347, opened 2026-02-02
(zune-jpeg) tweak truncated file behavior to help with issue #273

fpgaminer:
This pull request helps fix one of the images in issue #273 .

Image `582121.11.450x300.jpg` in that issue is a truncated JPEG.  jpeg-turbo will decode the image into something reasonable, whereas zune-jpeg puts a noise pattern over the image.

This patch does a small change to zune-jpeg's behavior here to eliminate the noise pattern.  This reduces the MAE between zune-jpeg and jpeg-turbo from ~8 to ~1.

## MAJOR NOTES

1) `582121.11.450x300.jpg` is a _broken_ image.  Normally I wouldn't care much how a library behaves on broken images.  But this patch is very small, and seems to improve zune-jpeg's behavior in these extreme edge cases.  So I figure it's worth it?

2) I ran a sweep across 10k random images on my end to make sure this patch doesn't affect anything else.  I didn't notice any regressions.

3) I ran a sweep across the images in test-images/jpeg and noticed **one regression**.  https://github.com/etemesi254/zune-image/blob/dev/test-images/jpeg/incomplete_image.jpg  It changes the color of the truncated blocks on that image from grey to a cream color that matches the rest of the image.  I actually think that _looks_ better, but it is a divergence from jpeg-turbo's behavior here (which outputs grey for those blocks).

## Minor Notes

1) cargo test passed on my machine
2) cargo fmt (nightly) did not throw any warnings on the lines of this patch

### DIFF
--- a/crates/zune-jpeg/src/bitstream.rs
+++ b/crates/zune-jpeg/src/bitstream.rs
@@ -187,8 +187,17 @@ impl BitStream {
         macro_rules! refill {
             ($buffer:expr,$byte:expr,$bits_left:expr) => {
                 // read a byte from the stream
-                $byte = u64::from(reader.read_u8());
-                self.overread_by += usize::from(reader.eof()?);
+                match reader.read_u8_err() {
+                    Ok(b) => {
+                        $byte = u64::from(b);
+                    }
+                    Err(_) => {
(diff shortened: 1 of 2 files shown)


fpgaminer:
Nevermind, need to work on this more, sorry.

Issue #351, opened 2026-02-09
Allow custom bit depths


feliwir:
Currently the `BitDepth` only allows `Eight` or `Sixteen` is integer bitdepth. However some formats allow for other bit depths inbetween (e.g. PPM or JPEG).  Currently it's not possible to encode / decode anything inbetween.

Issue #349, opened 2026-02-08
Support all JPEG transfer-syntaxes used by the DICOM standard


feliwir:
Hello,
thank you for this excellent library!
I'm using it to decode pixeldata contained in medical imaging, called the **DICOM** format (see https://www.dicomstandard.org/)
Unfortunately due to the nature of medical imaging it sometimes has unusual requirements, e.g. lossless jpeg, higher bit depth etc. which are not supported by this library yet (see https://www.dicomlibrary.com/dicom/transfer-syntax/)

Recently also JPEG-XL was added as a valid DICOM pixeldata format: https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.15.html

I'd like to create a table to track the progress of support & implementation
| Name                                 | Transfer Syntax              |  Decoding  | Encoding  |  Issue / PR | 
|------------------------------------|------------------------------------|----------------|-----------------|----------------|
|  JPEG P1                             | 1.2.840.10008.1.2.4.50  |       Yes        | Yes             |                    |
|  JPEG P2/4                         | 1.2.840.10008.1.2.4.51   |      No         | No              |   #353        |
|  JPEG P14                          | 1.2.840.10008.1.2.4.57   |      No         | No              |                    |
|  JPEG Lossless                  | 1.2.840.10008.1.2.4.70   |      No         | No              |    #144       |
| JPEG-LS (Lossless)           | 1.2.840.10008.1.2.4.80   |      No         | No              |    #144       |
| JPEG-LS (Lossy)                | 1.2.840.10008.1.2.4.80   |      No         | No              |    #144       |
|  JPEG 2000 (Lossless)      | 1.2.840.10008.1.2.4.90   |      No         | No              |                    |
|  JPEG 2000                        | 1.2.840.10008.1.2.4.91   |      No         | No              |                    |
| JPEG XL Lossless             | 1.2.840.10008.1.2.4.110 |       Yes        | Yes             |                    |
| JPEG XL Recompression| 1.2.840.10008.1.2.4.111 |       Yes        | No              |   #355        |
| JPEG XL                             | 1.2.840.10008.1.2.4.112 |       Yes        | Yes             |                    |

I will continue the table when i work on more things

etemesi254:
Hi, thank you for this.

Will you be working on it?

feliwir:
@etemesi254 yes, i am working on it right now. The extended JPEG decoding is already working (PR will land soon).
[...]

PR #352, opened 2026-02-09
core: Add custom bit depth field

feliwir:
This change allows specifying custom bit depths as specified in #351 
By itself this change allows to encode 10bit & 12bit PPMs which is a requirement for testing changes proposed in #349 

E.g. i can save the decoded 10bit & 12bit JPEG files now (PR will arrive after this one):
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,7 +404,7 @@ checksum = "e6506c6c10786659413faa717ceebcb8f70731c0a60cbae39795fdf114519c1a"
 dependencies = [
  "bytemuck",
  "byteorder-lite",
- "moxcms",
+ "moxcms 0.7.10",
  "num-traits",
  "zune-core 0.5.0",
  "zune-jpeg 0.5.6",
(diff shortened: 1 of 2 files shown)


etemesi254:
This has to be approached with care, the depth is important in other places especially in the imageprocs function, so will spend some few days checking that it works.

But appreciate the effort in starting it.

feliwir:
I choose this approach to stay API compatible. I'd prefer not having an enum at all, but i think that would be a breaking change.

etemesi254:
how would you represent different depths? disregarding apis?

feliwir:
Well, the bit depth would just be an integer (u8) without an enum in my opinion. Because an amount of bits is always a variable number and there might be anything really from 0-32. Maybe we have 64bits depths in the future :)

Currently the API limits us to three possible presets (8, 16, 32). Additionally it also mixes in the scalar type (float or integer). This is not directly related to the bit depth in my opinion

etemesi254:
And what datatype would represent them?

Because at the end of the day you need to tie it to a number a computer can manipulate.

feliwir:
I totally agree that a datatype needs to be selected. However the current API is misleading, because there are also 8 & 16bit floats, and there are also 32-bit integers.
So in my imagination there should be a way to select the data type (signed integer, unsigned integer or float)  and a function to select the bitdepth (a arbitary number). In case of an unsupported / invalid combination the encoder should notify the user by returning an error

etemesi254:
ignore the encoding for now

how would you perform operations ?

a simple operation is like increasing exposure so its usually 'clamp(num*exposure_value,MAX_SIZE,VAL)'

depth of 'f32' represents floating type images which are between 0-1 generally not MIN_FLOAT-MAX_FLOAT

feliwir:
I know that float image data is usually between 0-1. I am not sure how that is related to the API change proposal though

etemesi254:
It is unrelated, just trying to pick your brain on what a better design would be, maybe one day could add it.

PR #353, opened 2026-02-09
jpeg: Extended jpeg support

feliwir:
This builds on #352 

This adds support for process 4 of the JPEG specification (12-bit support). The decoding routines are similar to the existing 8-bit decoding routines, however i made the bitdepth a generic parameter. In theory everything in `mcu_ext.rs` could be applied to the functions in `mcu.rs` and then we could remove the duplicate code

For testing i added some publicly available DICOM files, provided by the standard committe. These files are publicly available here: ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG04/
It contains both extended jpeg (JPLY) files and lossless jpeg (JPLSS)

To access the pixeldata from these files i added https://github.com/Enet4/dicom-rs as a dev dependency

- [x] Profile if the generic variants of the scalar IDCT calculation are equally fast and replace the non-generic variants
- [ ] Stop generating PPM image files in the tests. How to validate correct decoding?


feliwir:
I made the IDCT functions generic. I think for the other functions we could do the same, but i am not sure how much generic usage you're a fan of

crates/zune-jpeg/Cargo.toml
```diff
@@ -29,8 +29,10 @@ default = ["x86", "neon", "std"]
 
 
 [dependencies]
+bytemuck = "1.25.0"
```
etemesi254: any reason for the new dependency?

crates/zune-jpeg/src/decoder.rs
```diff
@@ -785,7 +789,19 @@ where
         let out_len = core::cmp::min(out.len(), expected_size);
         let out = &mut out[0..out_len];
 
-        if self.is_progressive {
+        if self.is_extended {
+            let out_u16 = unsafe {
+                core::slice::from_raw_parts_mut(out.as_mut_ptr() as *mut u16, out.len() / 2)
```
etemesi254: Prefer `align_to` here.

crates/zune-core/src/bit_depth.rs
```diff
@@ -41,7 +41,15 @@ pub enum BitDepth {
...
-    Unknown
+    Unknown,
+    /// Custom integer bit depth, with the number of bits specified in the variant
+    ///
+    /// This is used for images with bit depths that do not fit into the standard 8 or 16 bit categories, such as 10-bit images.
+    /// The number of bits is specified in the variant.
+    ///
+    /// Data is stored in the next higher byte size, e.g. a 10-bit image
+    /// would be stored in a u16, and the valid range of values would be 0 to 1023.
+    Custom(u8)
```
Enet4, line 52: A bit of a shame that this will bump the size of the enum, even though in practice we wouldn't need all of the value range in a `u8` (some values would be  unacceptable and others not practically useful). But I'm not sure if there's a way around it other than expanding all other bit depths in-between, which isn't fun either. Another option is to give up on BitDepth as an enum and use a newtype around `u8`. Of course, this depends on the maintainer's willingness to the change.

crates/zune-jpeg/tests/dicom.rs
```diff
@@ -0,0 +1,203 @@
...
+        _ => {
+            panic!("Unsupported PixelData format: expected PixelSequence or Primitive")
+        }
+    }
+}
+
+/// Test decoding a 12-bit lossy JPEG image stored in a DICOM file.
+#[test]
+fn decode_dicom_lossy_extended_12bit() {
+    let test_data = include_bytes!("../../../test-images/jpeg/dicom/JPLY/MG1_JPLY");
```
Enet4, line 128: You can pluck the JPEG files out of the DICOM files using `dicom-toimage --unwrap`, so that you don't need to add `dicom-object` as a dependency.

PR #357, opened 2026-02-14
Fix a potential arithmetic overflow

LaurenzV:
I encountered this in an image with a very large height (due to erroneous metadata).


etemesi254:
Thanks, seems like its the only of such code.

Appreciate

PR #358, opened 2026-02-14
fix: progressive JPEG with DRI + 4:2:0 decoded as grayscale

lilith:
## Summary
- Fix `handle_rst()` erroring on valid between-scan markers (SOS, DHT, etc.) found at restart boundaries, which caused progressive 4:2:0 JPEGs with DRI to decode as grayscale
- Explicitly accept valid inter-scan markers at restart boundaries; unknown markers warn by default and error in strict mode (matching libjpeg-turbo behavior)
- Add regression test with embedded 479-byte JPEG that triggers the exact conditions

## Root cause

When the restart interval aligns with the end of a non-interleaved scan (e.g. 4 Y blocks at 4:2:0 with DRI=4), `handle_rst_main` fires at `todo == 0`. The bitstream `refill()` reads the next SOS marker from the stream. `handle_rst()` then hit its catch-all `_ =>` branch and returned `MCUError("Marker SOS found in bitstream, possibly corrupt jpeg")`. In non-strict mode this error was caught, logged, and `finish_progressive_decoding` was called immediately — with only Y DC coefficients filled, producing R==G==B grayscale output.

All three conditions are required to trigger: progressive (SOF2) + DRI present + chroma subsampling (4:2:0). Remove any one and decoding is correct.

## Test plan
- [x] New test `test_progressive_dri_420_color` with embedded JPEG passes
- [x] All existing zune-jpeg tests pass (23 tests)
- [x] Full workspace test suite passes (169 tests)


etemesi254:
Thank you for this

appreciated

PR #344, opened 2026-01-18
Add unsafe to functions in avx2.rs

AdeptVeritatis:
https://github.com/etemesi254/zune-image/issues/343

This fixes error[E0658] with Rust version 1.85.0.

PR #365, opened 2026-03-04
fix: unsafe_op_in_unsafe_fn compliance for NEON + CI aarch64 coverage

kubijo:
## Problem

Multiple `unsafe fn` functions across the codebase have bodies that call unsafe operations without explicit `unsafe {}` blocks. Under `unsafe_op_in_unsafe_fn` (deny-by-default in Rust edition 2024), these fail to compile.

The NEON upsampler functions additionally used `#[target_feature(enable = "neon")]` without being declared `unsafe fn`, which is the same class of issue fixed for AVX2 in #344.

There was also no CI coverage for aarch64 at the MSRV boundary, so these issues went undetected.

### CI (`f87b5bda`)
- Add `workflow_dispatch` trigger to the Test workflow
- Add `ubuntu-24.04-arm` / toolchain `1.88` matrix entry with `-D unsafe_op_in_unsafe_fn` in RUSTFLAGS
- **Failing run (before fix):** https://github.com/kubijo/zune-image/actions/runs/22755012242
- **Passing run (after fix):** https://github.com/kubijo/zune-image/actions/runs/22756419888

### Code (`d10991c3`)

**`zune-jpeg`** — NEON upsampler, color convert, IDCT, and utility files:
- `upsampler/neon.rs`: `pub fn` → `pub unsafe fn` for `upsample_horizontal_neon`, `upsample_vertical_neon`, `upsample_hv_neon`; add `unsafe {}` block in `upsample_hv_neon` body
- `color_convert/neon64.rs`: wrap `ycbcr_to_rgb_baseline_no_clamp` body in `unsafe {}`
- `idct/neon.rs`: wrap `idct_neon`, `condense_bottom_16`, `clamp256_neon` bodies in `unsafe {}`
- `unsafe_utils_neon.rs`: wrap `loadu`, `YmmRegister::load`, `transpose4`, `transpose` bodies in `unsafe {}`

**`zune-image`** — channel allocator:
- `channel.rs`: wrap unsafe ops in `alloc`, `realloc`, `dealloc`, `extend_unchecked`, `reinterpret_as_unchecked`, `alias`, `alias_mut`

**`zune-capi`** — C FFI:
- `utils.rs`: wrap unsafe ops in `zil_guess_format`, `zil_malloc`, `zil_free`, `zil_free_and_null`
- `errno.rs`: wrap unsafe ops in `zil_status_new`

## Note

The three nightly CI failures are a pre-existing `portable_simd` API breakage (`Mask::select` removed), unrelated to this PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,7 @@ on:
     branches: [ "main","dev" ]
   pull_request:
     branches: [ "main","dev" ]
+  workflow_dispatch:
 
 permissions:
   contents: read
@@ -28,6 +29,9 @@ jobs:
         include:
           - os: ubuntu-latest
             toolchain: 1.88 # MSRV
+          - os: ubuntu-24.04-arm
+            toolchain: 1.88 # MSRV (aarch64 — catches NEON compilation issues)
+            extra_rustflags: -D unsafe_op_in_unsafe_fn
           - os: windows-latest
             toolchain: stable
             test-args: --exclude zune-benches --exclude zune-image --exclude zune-imageprocs --exclude zune-png
@@ -64,4 +68,4 @@ jobs:
                 ${{ matrix.test-args }}
         env:
           # Skip target-cpu=native for cross-compilation targets
-          RUSTFLAGS: ${{ matrix.target && '' || '-C target-cpu=native' }}
+          RUSTFLAGS: ${{ matrix.target && '' || '-C target-cpu=native' }} ${{ matrix.extra_rustflags }}
--- a/crates/zune-capi/src/errno.rs
+++ b/crates/zune-capi/src/errno.rs
@@ -104,23 +104,24 @@ pub extern "C" fn zil_status_ok(status: *const ZStatus) -> bool {
 /// Remember to free it with `zil_status_free`
 #[no_mangle]
 pub unsafe extern "C" fn zil_status_new() -> *mut ZStatus {
-    let ptr = zil_malloc(size_of::<ZStatus>());
-
-    let msg = CString::new("").unwrap();
-    let mem = unsafe { zil_malloc(msg.as_bytes_with_nul().len()) };
-    if mem.is_null() {
-        return ptr::null_mut();
-    }
-    // copy to memory
     unsafe {
+        let ptr = zil_malloc(size_of::<ZStatus>());
+
+        let msg = CString::new("").unwrap();
+        let mem = zil_malloc(msg.as_bytes_with_nul().len());
+        if mem.is_null() {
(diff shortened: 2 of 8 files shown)


etemesi254:
Ahh, my mistake sorry.

How can we add it in CI?

kubijo:
I'm not sure there ... I have not investigated that setup well enough to say how to mitigate it there... Is that something you'd want me to boot on top of this fix as well?

etemesi254:
Yes,  would be very welcome.

There was this part that added msrv as 1.88, we can add another matrix to 1.85  and compile for neon there to run the tests preventing future regressions

https://github.com/etemesi254/zune-image/blob/05a21489ffd6de1e629ce89f51f5cb4d66dd3f0d/.github/workflows/test.yml#L25-L42

kubijo:
I made the changes and it uncovered quite a few new locations of the same error.
The PR description is updated.

etemesi254:
Thanks.

Is the unsafe function definition no longer working? 

Because i see most of it is `unsafe fn (){ unsafe{}}`

kubijo:
The `unsafe fn() { unsafe {} }` is what the `unsafe_op_in_unsafe_fn` lint requires — deny-by-default in edition 2024.

Old behavior: `unsafe fn` made the whole body an implicit unsafe context.
New behavior: `unsafe fn` only affects callers, the body needs explicit `unsafe {}` around actual unsafe ops. Better for auditing which operations are actually unsafe.

- Edition guide: https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html
- Lint docs: https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unsafe-op-in-unsafe-fn

The `-D unsafe_op_in_unsafe_fn` flag is now in the aarch64 MSRV CI job so this gets caught going forward. Nightly fails are pre-existing `portable_simd` breakage, unrelated.

kubijo:
Thanks for the quick merge!

Quick question — what's your release cadence / policy for publishing to crates.io? We're currently pinned to a git rev downstream and would love to move back to a proper crates.io dependency once a release with this fix lands.

etemesi254:
trying to merge the other prs before release

hopefully in the next 24 hours

etemesi254:
sorry for delay

just left with fixing the fuzz crashes

kubijo:
> sorry for delay
> 
> just left with fixing the fuzz crashes

That's ok man, I'm sure this is just your side hustle ... Should I take it that it's now released to cargo then?

etemesi254:
https://crates.io/crates/zune-jpeg/0.5.13

PR #366, opened 2026-03-05
Fix integer overflow on coefficient substraction

1c3t3a:
This was crashing for a weird image. Wrapping add and wrapping sub mimics what libjpeg_turbo does for the image.

### DIFF
--- a/crates/zune-jpeg/src/bitstream.rs
+++ b/crates/zune-jpeg/src/bitstream.rs
@@ -48,6 +48,7 @@
 //! Knock yourself out.
 use alloc::format;
 use alloc::string::ToString;
+use zune_core::log::warn;
 use core::cmp::min;
 
 use zune_core::bytestream::{ZByteReaderTrait, ZReader};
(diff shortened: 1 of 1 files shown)


etemesi254:
Hi, thanks for this.

But before i merge could you share the image for further analysis. Just need to confirm that was the only reason it wasn't decoding

crates/zune-jpeg/src/bitstream.rs
```diff
@@ -651,11 +651,8 @@ impl BitStream {
                         break 'no_eob;
                     }
                 } else {
-                    if symbol != 1 {
-                        return Err(DecodeErrors::HuffmanDecode(
-                            "Bad Huffman code, corrupt JPEG?".to_string()
-                        ));
-                    }
+                    // libjpeg-turbo doesn't return an error here, it only warns
+                    // but we don't have a way to warn, so we just continue
```
etemesi254: We do have a way to warn, you can just use the warn! macro

1c3t3a:
I am super sorry. The image is proprietary and I cannot share it with you here. It's hard to craft it myself, it is a picture that had the high frequency components moved to the right. Looks like it was progressively encoded and then broken somehow.

I verified that this is the only issue with the image and that it decodes just like libjpeg_turbo now.

crates/zune-jpeg/src/bitstream.rs
1c3t3a: Good call!

PR #367, opened 2026-03-19
jpeg: Add mpf offset to ImageInfo

vigneshvg:
The multi_picture_information data may contain offsets in it that are relative to the start of the mpf data. So the data itself is not completely useful without the start offset.

Add a new field to ImageInfo that exposes the start offset.


vigneshvg:
Thank you!

vigneshvg:
Could you please make a patch release to include this change?

I have created https://github.com/etemesi254/zune-image/pull/368. Not sure if there is anything else to be done to make a release.

etemesi254:
Done

PR #373, opened 2026-04-11
jpeg: implement arithmetic coding

### LINKED ISSUE #318: JPEG: Support DAC (define-arithmetic-coding)
Apparently, GIMP supports exporting it: https://github.com/user-attachments/assets/1895633e-2ad7-46a4-97ae-8ac61e0ed700

libjpeg-turbo supports decoding it

Example image: https://gitlab.gnome.org/-/project/27706/uploads/f8579c344173dc26f8a2e1ef155cd695/P1080392-raw-P1080403-raw.jpg

Glycin report: https://gitlab.gnome.org/GNOME/glycin/-/issues/179

mstoeckl:
This PR would close https://github.com/etemesi254/zune-image/issues/318. It implements support for arithmetic decoding for the sequential and progressive lossy JPEG modes. (Not lossless mode, since that is even rarer and zune-jpeg doesn't implement it with any encoding type.)

The test images were made by me and are released in the public domain, and also offered under MIT OR Apache-2.0 OR Zlib . (I used libjpeg-turbo's `jpegtran` to create a few sequential and progressive variants of a base image that I made.)

Note: running `cargo +nightly fmt` produces some format changes in unrelated code, so I've tried to just include the format changes applying to my own code.

I've run a fuzzer against this code for a few hours, seeded with a few arithmetic-encoded images, and have only found what appear to be existing crashes.

The new `arith` feature for zune-jpeg is off by default.

Please let me know if you'd like any changes.


etemesi254:
> I've run a fuzzer against this code for a few hours, seeded with a few arithmetic-encoded images, and have only found what appear to be existing crashes.

please share those ones too

mstoeckl:
> please share those ones too

Done, in my most recent two merge requests. With those applied on top of this branch, I haven't found any further issues even after fuzzing for a while.

etemesi254:
Don't have any problem, with the PR, very okay with merging.

etemesi254:
Got a chance to look at it closely, some few questions.


How does it handle truncated streams?

```rust
   overread_by:         usize,
    /// True if we have seen end of image marker.
    /// Don't read anything after that.
    seen_eoi:            bool,
```
the above variables in `BitStreamArithmetic` are not referred anywhere except the functions from the trait which means they are always what they were initialized from

the `seen_eoi` flag is to prevent the decoder from reading any more bytes after it has seen the end of image marker (useful for stuff like ultra-hd images which stitch 2 images in one container, separated by the EOI marker)

mstoeckl:
> the above variables in BitStreamArithmetic are not referred anywhere except the functions from the trait which means they are always what they were initialized from

No, `seen_eoi` is set by `mcu.rs` when the EOI marker is hit.

> How does it handle truncated streams?

When any marker is hit (not just EOI), the decompressor is continually fed compressed zero bytes. (Every block of coefficients decoded requires a finite number of decisions, so no infinite loops should be possible.)

Section D.1.8:
> Any trailing zero bytes already written to the entropy-coded segment and not preceded by a X’FF’ may, optionally, be discarded.

BitStreamArithmetic handles the end of the stream as if the bytes following that might be a marker, so it also feeds zeros bytes to the decompressor until the current blocks of coefficients are complete. And then the code should error in `get_marker()` of mcu_prog.rs because no marker was found.

AlbertNashr:
@mstoeckl  Great news, many thanks for your efforts! I'd be happy to test. Testing would hopefully work for me if a version of loupe with DAC support gets available as a Debian package (ideally, the one that works in both _trixie_ and _forky_); not sure how far their progress is. If the feature is still off by default, consider turning it on; the goal would be to ease downstream packaging. They ask for a release of zune-jpeg supporting arithmetic-coded JPEGs.

PR #374, opened 2026-04-12
jpeg: fix component stride calculation

mstoeckl:
The component strides were previously computed using `mcu_x` values derived from the cumulative maximum sampling factors, instead of total maximum over all components, as required by ITU T-81 A.1.1.

For normal images, where the Y component has the greatest resolution, this was OK; but the logic failed with images where Y was subsampled and Cb or Cr was not. 

This also fixes a panic that I noticed while fuzzing. The new test image is a cleaned up version of the broken image that I produced using `jpegtran`; using the notation of ITU T-81 A.1.1, it has parameters:

* H_0 = 1, V_0 = 1 (since H_max=2, V_max=4, each sample is stretched to cover 2x4 pixels)
* H_1 = 1, V_1 = 1
* H_2 = 2, V_2 = 4 (each component sample maps to one output pixel)

This PR also includes a fix to make the upsampling scratch space account for all components when completing baseline and progressive decoding.


etemesi254:
Hi, thanks for this.

Has been a major source of bugs.

Okay if i merge this before the other bitstream one(this is smaller to analyze)?

mstoeckl:
> Okay if i merge this before the other bitstream one

Yes, no need to ask. The arithmetic coding bitstream MR may take a while to review, and rebasing it is no trouble.

PR #376, opened 2026-04-15
jpeg: fix panic with unusual component count and colorspace

mstoeckl:
This is a fix for a panic that I noticed while fuzzing the `zune-jpeg` crate. On the following test image, whose key properties are that it has a YCCK color space specified through the App14 tag, but only 2 components:

[ycck2.jpg](https://github.com/user-attachments/assets/8bfc5a3d-575c-4c2c-8d48-d00faa28f337)

`zune -i ycck2.jpg` would panic with the key parts of the error output being the following:
```
WARN  [zune_jpeg::misc] Treating YCCK colorspace as YCbCr as component length does not match

thread 'main' (94394) panicked at crates/zune-jpeg/src/mcu_prog.rs:102:40:
index out of bounds: the len is 2 but the index is 2
```

I'm not sure what the correct behavior should be on images like it, but I have applied a twofold fix:
1. Locally fixing decode_mcu_ycbcr_progressive, so that the component count used always matches the more reliable self.components.len().
2. Only converting YCCK to YCbCr if the image actually has 3 channels; 2 channel images should then always be interpreted as MultiBand in non-strict mode.


crates/zune-jpeg/src/misc.rs
etemesi254, line 1: I think its better if we keep the previous check, and then add another new nested check for those conditions

i.e image is YCCK and with 2 components, change it to multiband

mstoeckl: OK. I've also updated the code to reject misspecified YCCK images in strict mode, and ended up reorganizing the relevant lines a bit.

Issue #375, opened 2026-04-14
Support for resuming decoding after UnexpectedEof in JpegDecoder


telecos:
As a continuation of post #371 I have created this issue for tracking work related to this feature request. My initial proposal is to follow a similar approach that we follow to incorporate this functionality to BmpDecoder from image crate:

https://github.com/image-rs/image/issues/2696

The idea is that, instead of delivering the full functionality in a single PR, we split in smaller self-contained PRs that can be easier to review and also will allow to refine the direction in which we drive implementation.

My initial thought is to start from a minimal state machine that is able to resume from start of each state, and then improving it by incorporating better progress tracking and more fine-grained states if required. 

At the moment I am working on first PR here:

https://github.com/telecos/zune-image/tree/feat/incremental-decode

I will publish PR and update this tracking issue once I have it ready for review.

PR #378, opened 2026-04-15
Introduce minimal state machine for resumable decoding in JpegDecoder

telecos:
Add minimal state machine towards incremental/resumable decoding support in zune-jpeg. This minimal state machine will track header vs scan states and resume from start of the state in case of recoverable error. Subsequent PRs will add granularity on the resumable capability to allow a fine-grained status tracking to allow resuming from more acurate possition.

When decoding from a stream that doesn't have all bytes available yet (e.g. network download), decode_headers() can now be retried after feeding more data. If it fails because the input ran out, the decoder resets cleanly so the caller can try again.

Changes:

- Add DecodingState enum to track whether the decoder is parsing headers or ready to decode scan data
- Add is_recoverable_eof() on DecodeErrors to distinguish "need more data" from real format errors
- Add reset_header_state() to cleanly reset all decoder fields on recoverable EOF
- Refactor decode_into() to delegate header parsing to decode_headers() instead of duplicating its logic
- Add incremental decoding tests (full decode, chunked headers, byte-at-a-time feeding)
- No breaking API changes. Existing callers are unaffected.

This work is part of #375

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -80,7 +80,7 @@ pub(crate) enum DecodingState {
     /// Headers not yet fully decoded.
     DecodeHeaders,
     /// Headers decoded; scan data starts at `scan_start_position`.
-    DecodeScan { scan_start_position: u64 },
+    DecodeScan { scan_start_position: usize },
 }
 
 /// An encapsulation of an ICC chunk
@@ -570,7 +570,8 @@ where
 
                         // Transition to scan phase, remembering where scan data starts.
                         if self.state == DecodingState::DecodeHeaders {
-                            let pos = self.stream.position()?;
+                            #[allow(clippy::cast_possible_truncation)]
+                            let pos = self.stream.position()? as usize;
                             self.state = DecodingState::DecodeScan {
                                 scan_start_position: pos,
                             };
@@ -918,7 +919,7 @@ where
             }
             DecodingState::DecodeScan { scan_start_position } => {
                 // Seek back to scan start (needed for retry after more data arrived).
-                self.stream.set_position(scan_start_position as usize)?;
+                self.stream.set_position(scan_start_position)?;
             }
         }
 
@@ -985,8 +986,8 @@ where
     fn reset_header_state(&mut self) {
         self.info = ImageInfo::default();
         self.qt_tables = [None, None, None, None];
-        self.dc_huffman_tables = [None, None, None, None];
-        self.ac_huffman_tables = [None, None, None, None];
+        self.entropy_tables.dc_huffman = [None, None, None, None];
+        self.entropy_tables.ac_huffman = [None, None, None, None];
         self.components.clear();
         self.h_max = 1;
         self.v_max = 1;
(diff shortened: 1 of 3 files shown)


etemesi254:
Hi

so if i understand is this to achieve something like network decoding?

telecos:
> Hi
> 
> so if i understand is this to achieve something like network decoding?

Yes, this is the main motivation of this feature request. In the case of Chromium, it is required to start rendering images even before image is completely received, and decoding should be able to complete at next call to `decode` method once data is completely received.

etemesi254:
That makes sense


I had once thought about that

and hence why i rewrote the io layer my idea was the io can do the waiting so what happens is when a decoder asks for more data the io reader lacks it,it waits for sometime (lets say download a but more data) 

but also this makes sense


Let me ask though what would make it work for lets say all the other libraries (e.g zune-bmp) from a high level? what changes are expected

telecos:
I think that change can be split into two layers:

- Reusable: Add `is_recoverable_eof()` to `ZByteIoError` and document the retry contract — when a read returns `NotEnoughBytes`, the stream position is already rewound, so the caller can provide more data and retry. Each codec's error type would just delegate to this. We may want to consider if we support a configuration for blocking vs non-blocking (despite blocking one can be easily built on top of the non-blocking)
- Codec-specific: The state machine itself is tightly coupled to each format's structure (JPEG has markers → SOS → entropy data, BMP has a fixed header → pixel rows, PNG has typed chunks, etc.), so each codec would need its own. The complexity varies — BMP would be straightforward, JPEG is the most involved.

etemesi254:
> I think that change can be split into two layers:
>Reusable: Add is_recoverable_eof() to ZByteIoError and document the retry contract — when a read returns NotEnoughBytes, the stream position is already rewound, so the caller can provide more data and retry. Each codec's error type would just delegate to this. We may want to consider if we support a configuration for blocking vs non-blocking (despite blocking one can be easily built on top of the non-blocking)
>Codec-specific: The state machine itself is tightly coupled to each format's structure (JPEG has markers → SOS → entropy data, BMP has a fixed header → pixel rows, PNG has typed chunks, etc.), so each codec would need its own. The complexity varies — BMP would be straightforward, JPEG is the most involved.

Definitely sounds better, 100% prefer this path

telecos:
> Definitely sounds better, 100% prefer this path

I created a separated PR with the reusable part of this PR:

https://github.com/etemesi254/zune-image/pull/380

telecos:
I rebased this branch to get changes from PR #380 . I think we also need to consider zune-jpeg's ExhaustedData decoded error as recoverable.

etemesi254:
Is this PR ready?

telecos:
> Is this PR ready?

Yes, this is the first PR towards resumable decoding. My plan is to continue with follow up smaller PRs which should be easier to review (I think 4 or 5, each of them self-contained so that it can be completed without breaking things).

etemesi254:
Apologies for delay in review then,  the changes seem acceptable, please rebase them on the main branch and I will merge them

telecos:
> Apologies for delay in review then, the changes seem acceptable, please rebase them on the main branch and I will merge them

I rebased but there are 2 failing CI after that. I proposed #387 to fix them.

PR #380, opened 2026-04-20
Add is_recoverable_eof() to ZByteIoError for resumable decoding support

telecos:
Add a method on ZByteIoError to detect errors caused by insufficient data (as opposed to format/corruption errors). This enables incremental/streaming decoders to distinguish 'need more bytes' from 'data is invalid'.

Returns true for:
- NotEnoughBytes: the reader requested more bytes than available
- StdIoError with UnexpectedEof (std feature only)

Includes unit tests for all variants.

This work is part of #375


etemesi254:
LGTM

PR #381, opened 2026-04-25
AVX2+NEON private API soundness fix

Shnatsel:
`ycbcr_to_rgb_avx2` function is safe to call, but does not perform a bounds check on the output buffer. The _rgba variant performs this check, as do all the scalar variants, but it was missed here.

`ycbcr_to_rgb_neon` and `ycbcr_to_rgba_neon` are both missing the check.

Due to the missed checks the  affected functions also do not honor the `offset` parameter, but since it's always 0 in practice that didn't matter either.

This is **not** a security issue in practice because the API is private and is always called with the appropriate buffer size and offset 0.

I have not benchmarked how this affects performance. If this is at all noticeable, we might want to make a larger refactoring by turning `out` into a `&mut [u8]` and removing the `offset` argument instead of the change proposed here.

### DIFF
--- a/crates/zune-jpeg/src/color_convert/avx.rs
+++ b/crates/zune-jpeg/src/color_convert/avx.rs
@@ -299,7 +299,7 @@ const fn shuffle(z: i32, y: i32, x: i32, w: i32) -> i32 {
     (z << 6) | (y << 4) | (x << 2) | w
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "std"))]
 mod safety_tests {
     use super::*;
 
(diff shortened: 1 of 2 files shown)


etemesi254:
Makes sense

i never wrote that and may have slipped out of my radar when reviewing it

Shnatsel:
Thank you!

I still want to follow up on this with more thorough benchmarking. If there's any slowdown I'll try to refactor it to reduce performance impact.

etemesi254:
I have no problem sacrificing speed for safety

and the slowdown should be minuscule at best

Issue #379, opened 2026-04-16
Support for raw component plane output (skip upsampling + color conversion) in JpegDecoder


telecos:
As a continuation of #371, I have created this issue to track a feature request for raw plane output — equivalent to libjpeg-turbo's `raw_data_out` mode.

The idea is to return decoded component planes directly after IDCT, without upsampling or color conversion. For a YCbCr 4:2:0 image this gives Y at full resolution, Cb and Cr at half resolution in both dimensions, with plane dimensions rounded up to DCT-block boundaries (multiples of 8).

This is used in Chromium/Skia for GPU-accelerated JPEG decode: the raw YCbCr planes are uploaded as separate textures and color conversion happens on the GPU. It is also useful for video pipelines (V4L2, VA-API) that consume YCbCr planes directly.

The decoder already has clean separation between IDCT and the upsample + color convert stages, so the feature should be a bypass branch in `post_process()` that copies each component's post-IDCT data directly to the output buffer.

I will start working on a first PR and update this issue once it is ready for review.

etemesi254:
thanks

does this return data in interleaved or raw formats?

and does libjpeg fill in the data row/plane wise?

telecos:
> does this return data in interleaved or raw formats?
> and does libjpeg fill in the data row/plane wise?

For the case of YUV (where [raw_data_out](https://skia.googlesource.com/skia/+/refs/heads/main/src/codec/SkJpegCodec.cpp#884) is set to true), the data is returned in **planar** format, not interleaved. The caller provides three separate buffers (one per component: Y, U, V), and libjpeg fills them all simultaneously on each call.

More specifically, [jpeg_read_raw_data](https://skia.googlesource.com/skia/+/refs/heads/main/src/codec/SkJpegCodec.cpp#944) works in MCU-row blocks. For 4:2:0 (v_samp_factor=2 for Y), each call delivers 16 rows of Y +
8 rows of U + 8 rows of V. The caller passes a `JSAMPIMAGE` — an array of three `JSAMPARRAY`s (one per component), each pointing to the corresponding rows in that component's plane buffer. After each call, all three planes advance together.

For comparison, the non-raw path uses [jpeg_read_scanlines](https://skia.googlesource.com/skia/+/refs/heads/main/src/codec/SkJpegCodec.cpp#450), which returns one row at a time of interleaved RGB/RGBA after libjpeg internally does upsampling + color conversion.

etemesi254:
[...]

PR #386, opened 2026-04-30
JpegDecoder: Add a borrowing raw component-output session

telecos:
## Summary

This is the first PR in the JPEG raw-output and scanline-output work.

It adds opt-in, whole-image raw component output through a borrowing `RawDecodeSession`.

Raw output returns one 8-bit post-IDCT plane per JPEG component, in SOF declaration order, without applying upsampling or color conversion. This supports consumers that process YCbCr, grayscale, RGB, CMYK, YCCK, or other component
planes themselves, including GPU upload workflows.

The existing public behavior of `decode()` and `decode_into()` remains unchanged.

## Public API

```rust
let mut decoder = JpegDecoder::new(ZCursor::new(data));
decoder.decode_headers()?;

let mut raw = decoder.raw_output();
let layout = raw.layout().unwrap();
let count = raw.num_components().unwrap();
let component_ids = raw.component_ids().unwrap();

let mut buffers: Vec<Vec<u8>> = (0..count)
    .map(|index| vec![0; layout[index].byte_size])
    .collect();
let mut planes: Vec<&mut [u8]> =
    buffers.iter_mut().map(Vec::as_mut_slice).collect();

raw.decode_into_planes(&mut planes)?;
```

Part of #379.

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -2137,6 +2137,9 @@ where
     ///
     /// # Examples
     ///
+    /// Decode a JPEG into raw YCbCr planes (libjpeg-turbo style) and access
+    /// individual component data:
+    ///
     /// ```no_run
     /// use zune_core::bytestream::ZCursor;
     /// use zune_jpeg::JpegDecoder;
@@ -2145,15 +2148,30 @@ where
     /// let mut decoder = JpegDecoder::new(ZCursor::new(&data));
     /// decoder.decode_headers().unwrap();
     ///
+    /// // Query the plane geometry (DCT-block-padded sizes).
     /// let layout = decoder.planar_layout().unwrap();
     /// let n = decoder.num_components().unwrap();
+    ///
+    /// // Allocate one buffer per component, sized to the full padded plane.
     /// let mut buffers: Vec<Vec<u8>> = (0..n)
     ///     .map(|i| vec![0u8; layout[i].byte_size])
     ///     .collect();
     /// let mut planes: Vec<&mut [u8]> = buffers.iter_mut().map(|b| b.as_mut_slice()).collect();
     ///
     /// decoder.decode_raw(&mut planes).unwrap();
-    /// // planes[0] = Y, planes[1] = Cb, planes[2] = Cr (for YCbCr)
+    ///
+    /// // For a 4:2:0 YCbCr image:
+    /// //   planes[0] = Y  (full resolution, padded to DCT blocks)
+    /// //   planes[1] = Cb (half resolution)
+    /// //   planes[2] = Cr (half resolution)
+    /// //
+    /// // Access a specific pixel's Y value:
+    /// let y_stride = layout[0].stride; // row pitch in bytes
+    /// let y_value = planes[0][/* row */ 10 * y_stride + /* col */ 20];
+    ///
+    /// // Identify component order for non-standard JPEGs:
+    /// let ids = decoder.component_ids().unwrap();
+    /// println!("Component order: {:?}", ids);
     /// ```
     ///
     /// # Errors
@@ -2246,8 +2264,10 @@ where
     ///
     /// # Examples
     ///
-    /// Skia-style: allocate with custom stride (e.g. GPU-aligned), decode
-    /// only the logical image area, leave padding untouched.
+    /// Skia-style: decode into GPU-aligned buffers where each row has a
+    /// power-of-two stride, then upload the planes as textures. Only the
+    /// logical image area is written; padding bytes are left untouched so
+    /// the caller can pre-fill them with a known value (e.g. for debugging).
     ///
     /// ```no_run
     /// use zune_core::bytestream::ZCursor;
@@ -2264,14 +2284,24 @@ where
     /// let strides: Vec<usize> = (0..n)
     ///     .map(|i| (layout[i].width + 63) & !63)
     ///     .collect();
+    ///
+    /// // Allocate buffers sized to logical height × custom stride.
+    /// // Only layout[i].width × layout[i].height pixels are written;
+    /// // the gap between width and stride is never touched.
     /// let mut buffers: Vec<Vec<u8>> = (0..n)
     ///     .map(|i| vec![0u8; strides[i] * layout[i].height])
     ///     .collect();
     /// let mut planes: Vec<&mut [u8]> = buffers.iter_mut().map(|b| b.as_mut_slice()).collect();
     ///
     /// decoder.decode_raw_strided(&mut planes, &strides).unwrap();
-    /// // Each planes[i] now contains the component data with the custom stride.
-    /// // Upload planes[i] to a GPU texture with row pitch = strides[i].
+    ///
+    /// // Upload each plane to a GPU texture:
+    /// for i in 0..n {
+    ///     let width = layout[i].width;
+    ///     let height = layout[i].height;
+    ///     let row_pitch = strides[i];
+    ///     // gpu.upload_texture(planes[i], width, height, row_pitch);
+    /// }
     /// ```
     ///
     /// # Errors
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -499,7 +499,14 @@ pub(crate) struct RawPlanesSink<'planes, 'buf> {
 /// skips upsampling and color conversion and returns one post-IDCT plane per
 /// JPEG component. The configured output colorspace is therefore ignored.
 pub struct RawDecodeSession<'decoder, T> {
-    decoder: &'decoder mut JpegDecoder<T>
+    decoder: &'decoder mut JpegDecoder<T>,
+    previous_incremental_mode: bool,
+}
+
+impl<T> Drop for RawDecodeSession<'_, T> {
+    fn drop(&mut self) {
+        self.decoder.incremental_mode = self.previous_incremental_mode;
+    }
 }
 
 impl<T> RawDecodeSession<'_, T>
@@ -874,6 +881,7 @@ where
         if let Some(state) = self.scan_state.as_deref_mut() {
             state.scan_checkpoint = None;
             state.progressive_checkpoint = None;
+            state.progressive_fine_checkpoint = None;
         }
     }
 
@@ -1190,7 +1198,12 @@ where
     /// pixel output or option changes until the session is dropped.
     pub fn raw_output(&mut self) -> RawDecodeSession<'_, T> {
         self.clear_scan_checkpoints();
-        RawDecodeSession { decoder: self }
+        let previous_incremental_mode = self.incremental_mode;
+        self.incremental_mode = true;
+        RawDecodeSession {
+            decoder: self,
+            previous_incremental_mode,
+        }
     }
 
     /// Number of components present in the JPEG scan (1..=4).
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -2133,6 +2133,27 @@ where
     /// meaningful; trailing padding columns / rows contain
     /// implementation-defined data.
     ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use zune_core::bytestream::ZCursor;
+    /// use zune_jpeg::JpegDecoder;
+    ///
+    /// let data = std::fs::read("photo.jpg").unwrap();
+    /// let mut decoder = JpegDecoder::new(ZCursor::new(&data));
+    /// decoder.decode_headers().unwrap();
+    ///
+    /// let layout = decoder.planar_layout().unwrap();
+    /// let n = decoder.num_components();
+    /// let mut buffers: Vec<Vec<u8>> = (0..n)
+    ///     .map(|i| vec![0u8; layout[i].byte_size])
+    ///     .collect();
+    /// let mut planes: Vec<&mut [u8]> = buffers.iter_mut().map(|b| b.as_mut_slice()).collect();
+    ///
+    /// decoder.decode_raw(&mut planes).unwrap();
+    /// // planes[0] = Y, planes[1] = Cb, planes[2] = Cr (for YCbCr)
+    /// ```
+    ///
     /// # Errors
     /// - [`DecodeErrors::TooSmallOutput`]: a plane buffer is shorter than
     ///   its `byte_size`.
@@ -2185,7 +2206,7 @@ where
         // Raw mode writes into the plane sink; downstream pixel bookkeeping
         // only needs a placeholder slice.
         let mut sink: [u8; 0] = [];
-        let result = if self.is_arithmetic {
+        if self.is_arithmetic {
             #[cfg(feature = "arith")]
             {
                 if self.is_progressive {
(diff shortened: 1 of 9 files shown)


etemesi254:
Thank you very much, 100% better than what I had.

> DecoderOptions::jpeg_set_raw_data_out 

Not sure i would want this in decoderoptions, my goal for decoderoptions was that it influences output we get from a `decode` function, so that e.g if we are making a command line utility, it makes sense to modify such (e.g to do direct ycbcr to grayscale, we can modify  `jpeg_set_out_colorspace`, to prevent malicious images with too many scans we can modify `jpeg_set_max_scans`)

`jpeg_set_raw_data_out` this influences **where** the output will be stored, diverging it to a different form, meaning some decode functions may break

E.g say zune-image has this.

```rust
    /// Open an encoded file for which the library has a configured decoder for it
    /// with the specified custom decoder options
    ///
    /// This allows you to modify decoding steps  where possible by specifying how the
    /// decoder should behave
    ///
    /// # Example
    ///  -  Decode a file with strict mode enabled and only expect images with less
    ///  than 100 pixels in width
    ///
    /// ```no_run
    /// use zune_core::options::DecoderOptions;
    /// use zune_image::image::Image;
    /// let options = DecoderOptions::default().set_strict_mode(true).set_max_width(100);
    /// let image = Image::open_with_options("/a/file.jpeg",options).unwrap();
    /// ```
    pub fn open_with_options<P: AsRef<Path>>(
        file: P, options: DecoderOptions,
    ) -> Result<Image, ImageErrors> {
        let reader = std::io::BufReader::new(std::fs::File::open(file)?);
        Self::read(reader, options)
    }
 ```
 So if i call it with 
 
 ```
 let options = DecoderOptions::default(). jpeg_set_raw_data_out(true)
let image = Image::open_with_options("/a/file.jpeg",options).unwrap();
 ```
 
 That wont work when decoding
 
 ## TLDR
 
 All that was to say, i would definirely prefer if instead of using `DecoderOptions` we had a flag in the `JpegDecoder` itself which controls raw mode

**Comment by etemesi254** [comment]
Also would be nice to seee full fledged examples,(later on before release, we would need to have docs explaining how to use that), definitely want to see how to use it skia style.

File: crates/zune-jpeg/src/decoder.rs [hunk 1]
```diff
@@ -42,9 +42,59 @@ use crate::upsampler::{
...
+    /// Allocated plane height in rows.
+    /// `ceil(height / 8) * 8`.
+    pub allocated_height: usize,
+    /// Required slice length for this plane: `stride * allocated_height`.
+    pub byte_size:        usize
+}
+
+/// Ceiling division: `ceil(a / b)` without overflow on the addition.
+#[inline]
+fn ceil_div(a: usize, b: usize) -> usize {
```
etemesi254: Won't this achieve the same ?

https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil

crates/zune-jpeg/src/decoder.rs
```diff
@@ -918,6 +1071,349 @@ where
...
+            #[cfg(not(feature = "arith"))]
+            unreachable!();
+        } else if self.is_progressive {
+            self.decode_mcu_ycbcr_progressive::<BitStreamHuffman>(&mut sink)
+        } else {
+            self.decode_mcu_ycbcr_baseline::<BitStreamHuffman>(&mut sink)
+        };
+
+        // Copy from internal buffers to the caller's output slices.
+        if let Some(raw_sink) = self.raw_planes_sink.take() {
```
etemesi254: How much work would it be to write directly to the caller output directly?

crates/zune-jpeg/src/decoder.rs
telecos: Right, good catch!

crates/zune-jpeg/src/decoder.rs
telecos: I reworked this to write directly into the caller-provided plane buffers.

I considered keeping the fully safe intermediate-buffer approach, but that adds both allocation and final copy-back overhead. The direct-write version keeps the API zero-copy by storing temporary pointers into the caller slices in `RawPlanesSink`, then clearing the sink before `decode_raw*` returns.

Because this requires a small amount of unsafe code, I changed `forbid(unsafe_code)` to `deny(unsafe_code)` so the unsafe is still rejected by default, and added localized `#[allow(unsafe_code)]` only for this raw output path.

crates/zune-jpeg/src/decoder.rs
```diff
@@ -1466,11 +1470,18 @@ where
...
-                    {
-                        dst[i] = *sample as u8;
+                debug_assert!(dst_offset + copy_w <= len);
+                // SAFETY: `ptr` points into the caller's `&mut [u8]` which is
+                // guaranteed to be at least `len` bytes.  The bounds check above
+                // (via `target_height` / `target_stride` validation in decode_raw*)
+                // ensures `dst_offset + copy_w <= len`.
+                unsafe {
+                    let dst = core::slice::from_raw_parts_mut(ptr.add(dst_offset), copy_w);
+                    for (i, sample) in src.iter().enumerate() {
```
etemesi254: Better if we use rust primitives here, removes the bounds check on `dst[i]`

```
 for (s,d) in src.iter().zip(dst.iter_mut()){
        *d = *s as u8;
    }
```
https://godbolt.org/z/WebxMdnvP

etemesi254:
I think a better way to handle the unsafe thing would be 

1. Change parameters on what `decode_mcu` takes in relation to the mode.

So maybe an enum

```rust
// Encapsulates the supported decoding types for MCU
enum McuDecodeType{
   Direct(& mut [u8]),
   Planes([& mut [u8];MAX_COMPONENTS]),
}
```

Do the normal decoding up until IDCT, perform IDCT. now on IDCT, depending on setting write to the component storage or a temporary data, convert to u8 and write to the plane.

```rust
// here  `out` assumes you are pointed to the decode type
 // this is the current code from mcu_rs:550
if component_samples_needed {
  // tmp was only written partially, note that len is in ZigZag order.
  clobber_more_than_4x4 = len > 10;

  let idct_position = if PROGRESSIVE {
      // For non-interleaved, j indexes data units directly
      j * 8
  } else {
      // derived from stb and rewritten for my tastes
      let c2 = v_samp * 8;
      let c3 = ((j * component.horizontal_sample) + h_samp) * 8;

      component.width_stride * c2 + c3
  };
  match out {
    McuDecodeType::Direct(pixels) => {
        // on direct, we have to buffer component data in our own channel, so use that 
        let idct_pos = channel.get_mut(idct_position..).unwrap();
        if len <= 1 {
            (self.idct_1x1_func)(tmp, idct_pos, component.width_stride);
        } else if len <= 10 {
            (self.idct_4x4_func)(tmp, idct_pos, component.width_stride);
        } else {
            //  call idct.
            (self.idct_func)(tmp, idct_pos, component.width_stride);
        }
      }
    }
    McuDecodeType::Planar(planes)=>{
      // for planar we write to a temp buffer of i16=>64, and then write on the stride as pure u8,
      // nb this must always be zero due to optimizations done on the IDCT params
      
      let temp_buf = [0_i16;64];
      // stride is by default 8, so that after this a whole 8x8 MCU is now stored in temp_buf
      if len <= 1 {
            (self.idct_1x1_func)(tmp, temp_buf,8);
        } else if len <= 10 {
            (self.idct_4x4_func)(tmp, temp_buf, 8);
        } else {
            //  call idct.
            (self.idct_func)(tmp, temp_buf, 8);
        }
        // now we are done, we need to write it to the output buffer, now do the strided access
        // TODO: confirm if id always between 0-4, not sure but this is a POC, so will do
        let out_stride_start = &mut planes[usize::from(component.id)][idct_pos..];
        // now slice and copy
        for (out_stride, tmp_chunk) in out_stride_start.chunks_exact_mut(component.width_stride).zip(temp_buf.chunks_exact(8)){
           // convert to u8
           let c:[u8;8] = core::array::from_fn(|c| tmp_chunk[c] as u8);
           // then copy
           out_stride[0..8].copy_from_slice(c);
        }
    }
}
```

The advantage of this is we can also remove the component allocations that are there, because depending on the output we can pass a flag to the part allocating the vecs saying if this is a planar output do not allocate

We also simplify the logic later on, e.g when we get to skipping post process, we just check what the caller has passed, whether its a `McuDecodeType::Direct` we call into post process, and if not, we yeild the execution

crates/zune-jpeg/src/decoder.rs
telecos: Good suggestion. Updated.

telecos:
I adopted the enum-based decode-mode plumbing from your suggestion: raw output is now passed through the MCU paths as `McuDecodeOutput::{Pixels, RawPlanes}` instead of being stored on `JpegDecoder`. That also let us remove the raw pointer sink, `unsafe` raw-plane writes, and the `Send`/`Sync` impls.

I kept the actual raw write at the MCU-stripe boundary for this patch, rather than moving it into the per-block IDCT branch. So the current version addresses the safety/API-shape part of the feedback, but it does not yet implement the allocation-skipping optimization where planar mode IDCTs into a temporary 8x8 block and writes directly to the destination plane.

I think that direct-IDCT-to-plane path is a good follow-up optimization, but it needs a more careful block-coordinate refactor across subsampling, non-interleaved scans, and progressive decode. I wanted this patch to stay bounded and keep the current raw-output tests stable.

etemesi254:
Hi, small update, cleaned up mcu_prog a bit so you may have to rebase before continuing,

Apologies for the inconveniences.

telecos:
> Hi, small update, cleaned up mcu_prog a bit so you may have to rebase before continuing,
> 
> Apologies for the inconveniences.

I’ve rebased the branch onto the latest dev and adapted the raw path to the current progressive MCU implementation.

This PR is now scoped as PR1: a borrowing `RawDecodeSession` for safe whole-image raw component output. It uses the shared `McuDecodeOutput` pipelines, writes directly to caller slices without new unsafe code, and preserves progressive, non-interleaved, arithmetic, retry, and replay behavior.

Planned follow-ups:

PR2: pull-based raw iMCU-row output equivalent to jpeg_read_raw_data
PR3: true converted scanline output equivalent to jpeg_read_scanlines

telecos:
@etemesi254, a quick update after splitting this work into smaller, stacked changes and validating the complete flow in Skia.

This PR is now limited to the first layer of the work:

- borrowing `RawDecodeSession` for whole-image raw component output;
- component IDs, sampling factors, and logical/padded plane geometry;
- padded and caller-strided destination planes;
- post-IDCT samples without upsampling or color conversion;
- recoverable input suspension, retry, and replay;
- no changes to the public behavior of `decode()` or `decode_into()`;
- no new unsafe raw-plane output code.

The remaining work is split into this PR and two dependent follow-ups:

1. #386: whole-image raw component output
2. #440: pull-based raw iMCU-row output, built on 386
3. #441: pull-based converted scanline output, built on 440

We have also integrated and exercised the complete three-PR stack in the corresponding Skia CL:

https://skia-review.googlesource.com/c/skia/+/1352756

That integration validates the intended Chromium/Skia use case, including raw YUVA output against the existing libjpeg-turbo codec, caller plane strides and padding, supported sampling modes, progressive and non-interleaved input, converted scanline reads/skips, and incomplete-input retry. It did not identify any missing zune-jpeg API required by the integration.

The later #440 / #441  optimization commits do not change the API validated by the Skia CL.

With the scope now reduced and the downstream integration validated, could you please review #386 when you have a chance? Once it lands, we can restack #440 on `dev`, followed by #441.

etemesi254:
Hi, sorry for delay.

Now does the api in the first comment https://github.com/etemesi254/zune-image/pull/386#issue-4358805940 still stand? or did it change as time went by

telecos:
> Hi, sorry for delay.
> 
> Now does the api in the first comment [#386 (comment)](https://github.com/etemesi254/zune-image/pull/386#issue-4358805940) still stand? or did it change as time went by

Hi, no problem. The overall design still stands, but the exact public API has evolved while I worked on integration with Chromium.

At commit 78795a026fd05e55adf440ea789543d8cc6d40de API was still exposed directly in `JpegDecoder`:

```
decoder.planar_layout()
decoder.decode_raw(&mut planes)
decoder.decode_raw_strided(&mut planes, &strides)
```

Internally, that revision had already adopted the enum-based suggestion

```
McuDecodeOutput::{Pixels, RawPlanes}
```

The output target was passed through the baseline and progressive MCU paths for each decode call. This removed decoder-stored raw pointers, unsafe plane writes, and the associated manual `Send` / `Sync` implementations.

What changed afterwards was mainly the public ownership model. The raw API is now grouped behind an exclusive borrowing session trying to use a wrong mixed use of regular API and new API:

```
let mut raw = decoder.raw_output();
let layout = raw.layout();
let ids = raw.component_ids();
raw.decode_into(&mut planes)?;
// or:
raw.decode_into_strided(&mut planes, &strides)?;
```

This change was made because raw decoding can suspend on incomplete input and resume on the same deocder. Holding `&mut JpegDecoder` in `RawDecodeSession` prevents callers from changing options or switching to pixel output in the middle of that sequence. It also keeps entropy, MCU, checkpoint, and replay state in one decoder rather than introducing a second decoding implementation. the session enables recoverable-input semantics while active and restores the decoder's previous mode when dropped.

We also extended the metadata contract. `PlaneInfo` now reports exact horizontal and vertical sampling factors, and `component_ids()` returns the actual SOF selector bytes rather than assuming positional `Y/Cb/Cr` identifiers.

The broader goal became clearer during implementation: we need both raw component output and a real pull-based scanline mechanism for the Skia/Chromium integration. Those have related internals but different public contracts, so we split them into three reviewable layers:

#386: borrowing session with whole-image raw component output.
#440: RawDecodeSession::decode_next_imcu_row, the pull-based equivalent of jpeg_read_raw_data, returning RowReady, NeedMoreInput, or Complete.
#441: ScanlineDecodeSession, with start, read_scanlines, skip_scanlines, and finish, analogous to converted jpeg_read_scanlines output.

This split also let us define the lifecycle carefully. Pull calls are sequential, decoder progress is retained across calls, invalid arguments do not claim ownership of a sequence, and NeedMoreInput does not publish partially committed output. Single-SOS baseline images can stream naturally, while progressive and multi-SOS images must first retain coefficients and then expose rows sequentially.

One part of the original sketch remains intentionally deferred: direct per-block IDCT into caller planes. The current implementation still does not IDCT into a temporary 8×8 block and write that block directly to the destination plane. Implementing that safely requires a deeper block-coordinate refactor covering subsampling, non-interleaved scans, progressive reconstruction, edge clipping, and retry boundaries. I kept it separate so this API work would not combine a correctness-sensitive hot-path rewrite with the ownership and streaming changes.

etemesi254:
Wait...

Does this change the api signature of "decode_into"?

telecos:
> Wait...
> 
> Does this change the api signature of "decode_into"?

No, the existing `JpegDecoder::decode_into` API signature does not change. It still accepts a single pixel-output buffer:

```
decoder.decode_into(&mut pixels)?;
```

The `decode_into` shown in my example is a new method on `RawDecodeSession`, not a replacement or overload of the existing `JpegDecoder` method:

```
let mut raw = decoder.raw_output();
raw.decode_into(&mut planes)?;
```

The implementation of the existing method now shares the internal output-mode plumbing, but its public signature remains unchanged.

telecos:
I renamed the new method to `decode_into_planes` to avoid future confusion with existing `decode_into`

etemesi254:
Ah, thanks, clears up some things,

No objections from my end.

Waiting on your last greenlight

telecos:
> Ah, thanks, clears up some things,
> 
> No objections from my end.
> 
> Waiting on your last greenlight

I think that this one is ready to merge. I am working on the aggregate of this one + #440 + #441 on Chromium side to complete all validations with no further issues found at the moment, so once this complete I will rebase and move next to ready to review.

PR #388, opened 2026-05-06
JpegDecoder: Add fine-grained resumable header parsing

telecos:
This PR adds fine-grained resumable JPEG header parsing. Instead of restarting header parsing from SOI after recoverable EOF, the decoder now checkpoints at the last fully consumed marker and resumes from that marker boundary when more input becomes available.

## Changes

- Add recoverable EOF classification for byte-stream errors.
- Track JPEG decode phase with `DecodingState::DecodeHeaders { resume_position }` and `DecodeScan { scan_start_position }`.
- Resume header parsing from the last fully consumed marker instead of reparsing previous markers.
- Roll back append-only metadata state when parsing an ICC, extended XMP, or gain-map marker fails partway through.
- Preserve scan retry behavior by seeking back to the recorded scan start position.
- Add incremental tests covering:
  - recoverable EOF on truncated input,
  - repeated EOF on the same decoder,
  - in-place growable stream header resume,
  - byte-by-byte fine-grained checkpointing,
  - decode parity with one-shot decoding.

This work is going towards completing #375

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -77,10 +77,45 @@ pub type IDCTPtr = fn(&mut [i32; 64], &mut [i16], usize);
 /// Tracks the current decoding phase for incremental decoding.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum DecodingState {
-    /// Headers not yet fully decoded.
-    DecodeHeaders,
+    /// Headers not yet fully decoded. `resume_position == 0` means start from SOI.
+    DecodeHeaders { resume_position: usize },
     /// Headers decoded; scan data starts at `scan_start_position`.
     DecodeScan { scan_start_position: usize },
+}
+
+// Snapshot of append-only buffers populated across multi-segment markers
+// (ICC chunks, extended XMP, gain map). All other parsers either overwrite
+// fixed slots (qt/huffman/components) or fail before mutating, so only
+// these three need to be rolled back when a marker parser errors out.
+#[derive(Clone, Copy)]
+struct HeaderAppendStateSnapshot {
+    icc:  usize,
+    xmp:  usize,
+    gain: usize
+}
+
+impl HeaderAppendStateSnapshot {
+    fn capture<T: ZByteReaderTrait>(decoder: &JpegDecoder<T>) -> Self {
+        Self {
+            icc:  decoder.icc_data.len(),
+            xmp:  decoder.extended_xmp_segments.len(),
+            gain: decoder.info.gain_map_info.len()
+        }
+    }
+
+    fn rollback<T: ZByteReaderTrait>(self, decoder: &mut JpegDecoder<T>) {
+        decoder.icc_data.truncate(self.icc);
+        decoder.extended_xmp_segments.truncate(self.xmp);
+        decoder.info.gain_map_info.truncate(self.gain);
+    }
+}
+
+/// Result of handling a single marker inside the header loop.
+enum MarkerStep {
+    /// Continue reading further markers.
+    Continue,
+    /// Reached SOS; headers are done and scan starts at the current position.
+    EnteredScan
 }
 
 /// An encapsulation of an ICC chunk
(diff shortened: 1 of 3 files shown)


etemesi254:
Hi, thanks for the PR.

Logically makes sense.

Though there is something else about jpegs where markers  are supported in the bitstream in non-strict mode.

This is to match libjpeg-turbo, ideally I have only encountered such images in fuzz tests (libjpeg-turbo would output an image where zune-jpeg would return an error), not sure about real world use cases.

The problem is now do these two mechanisms co-exist?

https://github.com/etemesi254/zune-image/blob/56ae0fc3b3de331622758291f3ae6aab4949023f/crates/zune-jpeg/src/mcu.rs#L594-L658

telecos:
> Hi, thanks for the PR.
> 
> Logically makes sense.
> 
> Though there is something else about jpegs where markers are supported in the bitstream in non-strict mode.
> 
> This is to match libjpeg-turbo, ideally I have only encountered such images in fuzz tests (libjpeg-turbo would output an image where zune-jpeg would return an error), not sure about real world use cases.
> 
> The problem is now do these two mechanisms co-exist?
> 
> https://github.com/etemesi254/zune-image/blob/56ae0fc3b3de331622758291f3ae6aab4949023f/crates/zune-jpeg/src/mcu.rs#L594-L658

Thanks for pointing to this gap. I will start adding a unit test covering that case which should fail in current implementation and I will try to come with some solution for supporting also this specific case in which we would have additional metadata outside of the `DecodeHeaders` state

telecos:
I looked into this deeper and here's where I landed:

For this PR I moved the snapshot/rollback logic inside `parse_marker_inner`` itself so every caller is protected uniformly — including the non-strict inline-marker dispatch from ``mcu.rs`. I also added an append-state snapshot at scan entry that gets rolled back before any retry seek, so scan replay can't duplicate ICC/XMP/gain-map entries even if it re-encounters the same inline markers.

While writing the regression test I found a deeper issue: scan-phase incremental resume doesn't actually work yet because the entropy decoder treats EOF as "scan finished" instead of signalling a recoverable error. So `decode_into` returns success with blank pixels and never gets a chance to retry. Fixing that requires changes inside `BitStream` + resetting MCU state on re-entry, which I think it is better to do as part of a separated PR which bring al the support for resumability during scan phase.

I will push the snapshot/rollback changes here together with the regression test (marked #[ignore] until the entropy decoder side is addressed). I also am thinking on creating a separate fuzz target covering the incremental/resumable (decode_incremental) that exercises the chunked feeding path — the existing fuzz corpus already has many file types that would be great to fuzz for the resumable decoding case.

etemesi254:
> While writing the regression test I found a deeper issue: scan-phase incremental resume doesn't actually work yet because the entropy decoder treats EOF as "scan finished" instead of signalling a recoverable error. So decode_into returns success with blank pixels and never gets a chance to retry. Fixing that requires changes inside BitStream + resetting MCU state on re-entry, which I think it is better to do as part of a separated PR which bring al the support for resumability during scan phase.

Yeah, but that does not fall under headers so we can deal with it when headers are complete.

We never lacked incremental decode support so truncated before meant a blank image, but when this is done, we will update the docs to show how one can do incremental :) 

> I will push the snapshot/rollback changes here together with the regression test (marked #[ignore] until the entropy decoder side is addressed). I also am thinking on creating a separate fuzz target covering the incremental/resumable (decode_incremental) that exercises the chunked feeding path — the existing fuzz corpus already has many file types that would be great to fuzz for the resumable decoding case.

Yes, please do

telecos:
> Yes, please do

I updated PR and created a separated PR for creating incremental decoding fuzzer in #389

etemesi254:
The PR looks good to me.

Will wait for your go ahead to merge in case you have any other thing you may wish to add

telecos:
Thanks! I did one final pass before giving the go-ahead and found a small header-resume edge case in the APP marker parsers.

APP1/APP2/APP13/APP14 were still using the non-error-returning byte reads for marker lengths / small metadata fields (`get_u16_be`, `read_u8`). With truncated incremental input, that could turn a temporary EOF into a non-recoverable format error, for example `Too small app2 segment`, instead of letting the caller feed more bytes and retry.

I pushed a small fix to use the error-returning reads there, and added a focused regression test: `header_app2_truncation_is_recoverable`. I also verified the regression fails without the fix (`Too small app2 segment` at the truncated APP2 marker) and passes with the fix.

I also started the work on supporting resumable scan decoding on a new branch based on this PR. At this point I think this PR is ready to complete, and we can switch the remaining scan-phase work to the follow-up PR after this merges, so I can rebase cleanly and open that separately.

197g:
This PR (specifically babf5c25fcf8c96e0379699c57284edfec41e1b3) is found by `git bisect` as the source of breaking `image`'s test suite for the jpeg integration of `zune-jpeg`, while I'm currently trying out the git tree as a pre-release. I'm not entirely certain if this might be the result of some necessary API / configuration change, any advice would be welcome.

etemesi254:
Hi,

which test case actually breaks?

197g:
This one here: https://github.com/image-rs/image/blob/44ce9226c541dd2b11897b9ef07156a28871cdb7/src/codecs/jpeg/decoder.rs#L245-L260

Curiously, converting the image with the converter example from `zune-image` still works. 
cat
 Does it have to do with us using a buffered file reader?

Also the result of this progressive image is no longer as expected: 

noop

telecos:
I have done a first quick analysis on both issues:

- I think that issue in `cat.jpg` is combination of new resumable decoding and lenient mode support: The truncated cat.jpg test now returns ExhaustedData in non-strict mode because resumable scan decoding propagates EOF before applying the previous lenient behavior. There is currently no configuration option that restores it.
- The corrupted output for 3.jpg was an unrelated bug introduced in [961f9324](https://github.com/etemesi254/zune-image/commit/961f9324581a1dd9859e2fe5df8846b864ecf652): progressive decoding selected the DC Huffman table using division instead of the original bitmask. Restoring the original table selection should fix the image I think.

So I think both cases will require fix from zune-jpeg side. For first one, if lenient mode is aimed to accept also truncated files, we may avoid having to add yet another configuration by documenting a limitation that lenient mode cannot work with resumable decoding, and do appropriate changes to treat EOF differently in lenient/non-strict model.

telecos:
and I think corrupted output issue is same issue as the one we just received in #430

telecos:
PR #431 should restore non-strict mode behaviour for truncated files.

PR #390, opened 2026-05-07
JpegDecoder: Add restart-boundary resumability for JPEG scan decoding

telecos:
This PR adds the first stage of resumable JPEG decoding support. It makes scan decoding retryable after recoverable EOF by preserving decoder state at scan start and, when restart markers are available, at the latest restart-marker boundary.

The goal of this PR is to support a practical and bounded resumability step without trying to match the full suspension model of libjpeg-turbo in one change.

What This Includes

- Tracks decoder phase across header and scan decoding.
- Allows header parsing and scan decoding to return recoverable EOF and retry on the same decoder after more input arrives.
- Rolls back append-only metadata collected during scan replay, so inline APP/COM marker handling does not duplicate ICC/XMP/gain-map state.
- Captures SOS-related state before scan replay, including component table selections and scan parameters.
- Adds restart-boundary checkpoints that preserve:
  - stream position after RST
  - MCU row/column resume point
  - DC/arithmetic restart state expectations
  - coefficient/sample buffers
  - component state
  - already-written output prefix
- Restores baseline, progressive, and arithmetic scan state from the latest checkpoint when possible.
- Keeps output validity tied to successful completion of `decode_into`; partial output after recoverable EOF remains retry state, not final decoded output.
- Adds integration coverage for incremental decoding across baseline, progressive, and arithmetic JPEGs.

Follow-Up Plan

- Follow-up PR 1: Preserve partial marker parsing state more closely, especially for standard marker payloads.
Avoid reprocessing incomplete marker bodies where libjpeg-turbo would suspend and continue.
Improve parity around APP/COM marker save behavior and input-controller state transitions.
- Follow-up PR 2: Explore preserving entropy decoder and MCU controller state inside restart intervals.
Reduce replay work when EOF happens mid-MCU or mid-restart segment.
Move closer to libjpeg-turbo’s coefficient-controller suspension behavior.

This work is part of #375

### DIFF
--- a/benchmarks/benches/decode_jpeg.rs
+++ b/benchmarks/benches/decode_jpeg.rs
@@ -243,6 +243,142 @@ fn decode_no_samp_opts(c: &mut Criterion) {
     });
 }
 
+// ---------------------------------------------------------------------------
+// Restart-marker / resumability path
+//
+// The benchmarks above all use JPEGs without restart markers (DRI), so the
+// per-RST checkpoint code added for incremental decoding is never exercised
+// by them. The benchmarks below close that gap with a DRI-bearing input
+// (four_components.jpg, restart interval = 7 MCUs, 4-component baseline).
+// They cover:
+//   * one-shot decode of a DRI image (steady-state RST checkpoint cost),
+//   * an incremental decode that triggers a recoverable-EOF retry, so the
+//     RST checkpoint capture + resume seek path is measured end-to-end.
+// ---------------------------------------------------------------------------
+
+use std::cell::Cell;
+use std::io::{BufRead, Read, Seek, SeekFrom};
+use std::rc::Rc;
+
+/// Byte-slice cursor with an externally-adjustable visibility limit.
+///
+/// Reads/seeks beyond `limit` behave as if the data ended there (EOF). The
+/// bench grows `limit` via the shared `Rc<Cell<usize>>` to simulate more
+/// data arriving on the same decoder.
+struct GrowableCursor<'a> {
+    data:     &'a [u8],
+    position: usize,
+    limit:    Rc<Cell<usize>>
+}
+
+impl<'a> GrowableCursor<'a> {
+    fn new(data: &'a [u8], limit: Rc<Cell<usize>>) -> Self {
+        Self { data, position: 0, limit }
+    }
+
+    fn visible(&self) -> usize {
+        self.limit.get().min(self.data.len())
+    }
+}
+
+impl Read for GrowableCursor<'_> {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        let visible = self.visible();
+        if self.position >= visible {
+            return Ok(0);
+        }
+        let available = &self.data[self.position..visible];
+        let n = available.len().min(buf.len());
+        buf[..n].copy_from_slice(&available[..n]);
+        self.position += n;
+        Ok(n)
+    }
+}
+
+impl BufRead for GrowableCursor<'_> {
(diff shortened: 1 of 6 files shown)


etemesi254:
Generally okay, appreciate the context, stuff, also was thinking the function parameters are too many.

A bit worried on the clones, especially where we have component buffers, 

most of the speed wins in this library have come from reducing allocation paths, initially in the first iterations of the library, the code used to first do entropy decoding, and then store that in one big array and then from that array do idct and post process (this was in baseline decode which you can decode per MCU row), initial benchmarks showed that as a massive bottleneck, i think a 15-20% difference was achieved when i switched to per MCU row decoding, even with its complications.

So if this now goes to Chromium which runs on diverse devices, we may have regressions just because of allocations, so thats why I am a bit pendantic on that.

telecos:
Thanks for the review - 

> Generally okay, appreciate the context, stuff, also was thinking the function parameters are too many.

RBundled the 8 individual MCU-decode parameters into a small `McuWidthContext` value. `decode_mcu_width` and `inner_decode_mcu_width` now take a single argument, and the `[#[allow(clippy::too_many_arguments)]` is removed. Inside the hot loop the context is destructured into local mutable bindings at entry so the optimizer sees plain &mut references.

> A bit worried on the clones, especially where we have component buffers,
[...] So if this now goes to Chromium which runs on diverse devices, we may have regressions just because of allocations

Thanks for pointing to this issue. The original change snapshotted full `Vec<i16>` coefficient buffers, a `Vec<Components>` and an output-prefix `Vec<u8>` into every `ScanCheckpoint`. On a DRI=7 image (~1000 RSTs per decode) that was ~900MB memcpy per decode. I have fixed this by:
- `ScanCheckpoint` is now allocation-free by construction. Only `Copy` scalars + a `[(i32, i32); MAX_COMPONENTS]` of DC predictor state.
- Coefficient buffers stay on the decoder.
- Progressive does not record per-RST checkpoint (refine passes do read-modify-write on `block`, so mid-scan resume would re-apply partial deltas). Progressive falls back to scan-start replay on EOF.
- `handle_rst_main` split into two variants so the per-MCU call sites in baseline and progressive don't share codegen budget.
- `DecodingState` collapsed into a flat `header_resume_position: usize` + `Option<Box<ScanDecodeState>>`. Removed the per call state clone.

> most of the speed wins in this library have come from reducing allocation paths, [...] initial benchmarks showed that as a massive bottleneck

To validate that new code does not reintroduce that bottleneck, I added a jpeg: DRI / Restart markers benchmark group (using four_compoenents.jpeg, DRI=7). And run benchmarks comparing this branch with dev `26797e76`:

Benchmark | dev (ms) | PR (ms) | Δ
-- | -- | -- | --
DRI / Restart markers — one-shot | 6.17 | 6.11 | −1.02%
No sampling Baseline | 82.76 | 81.88 | −1.06%
Horizontal Sub Sampling | 72.97 | 72.21 | −1.04%
Vertical sub sampling | 75.50 | 75.69 | +0.26%
HV sampling | 67.74 | 67.99 | +0.36%
Grayscale | 54.86 | 54.51 | −0.64%
No sampling Progressive | 276.79 | 275.76 | −0.37%
Progressive H | 221.30 | 221.49 | +0.09%
Progressive V | 222.09 | 220.40 | −0.76%
Progressive HV | 221.99 | 226.08 | +1.84%
Intrinsics (intrinsics) | 82.75 | 81.12 | −1.98%
Intrinsics (no intrinsics) | 82.67 | 81.25 | −1.72%

etemesi254:
Nice,

Some benchmarks on a hetzner server

https://github.com/etemesi254/zune-image/tree/inflate-new/benchmarks#jpeg-decoding

PR #391, opened 2026-05-09
JpegDecoder: Add libjpeg-turbo-style scanline output API

ender672:
Hello, I'm tracking an issue with large images in Gnome Nautilus (https://github.com/image-rs/image/issues/1989).

The code path is Nautilus -> glycin -> image-rs -> zune-image

In order to reduce memory use (and do high-performance scaling) I would like if zune-image returned one scanline at a time.

This PR adds a libjpeg-turbo-style scanline output API to `JpegDecoder`. It mirrors `jpeg_start_decompress` / `jpeg_read_scanlines` / `jpeg_finish_decompress`, giving callers a row-at-a-time output.

This is a first stage to explore the API. The first cut runs `decode()` at `start_decompress` and serves rows from the resulting buffer. The public contract matches a true streaming implementation, so a follow-up can swap in per-MCU-row decoding for the streamable cases (single-SOS baseline, progressive post-process) without changing the API or breaking callers.

What This Includes

- `start_decompress()`
- `read_scanlines(buf, num_lines) -> Result<usize>` (returns rows written)
- `finish_decompress()`
- `next_scanline()` (mirrors libjpeg's `cinfo.output_scanline`)
- New header-gated accessors returning `Option<usize>`, matching the existing `output_buffer_size` style:
  - `output_height` / `output_width` / `output_components` / `output_row_stride`
- libjpeg-turbo-compatible `read_scanlines` semantics:
  - oversize `num_lines` is capped to rows remaining (not an error)
  - reads past end return `Ok(0)`
  - the `start_decompress` precondition is checked before the `num_lines == 0` short-circuit, so the error is consistent regardless of buffer size
- New `crates/zune-jpeg/src/scanline.rs` carrying the per-call `ScanlineState`, populated by `start_decompress` and cleared by `finish_decompress`.

Tests

`crates/zune-jpeg/tests/scanlines.rs` cross-checks the scanline API against `decode()` across:
- common subsampling (4:2:0, 4:2:2, 4:4:4, horizontally-sampled, vertically-sampled, partial-MCU height)
- progressive (grayscale, 4:2:0)
- output-colorspace permutations (RGBA, luma-from-color, default RGB)
- multi-SOS baseline and baseline with restart markers (DRI)
- error and edge paths (`read_scanlines` before `start_decompress`, double `start_decompress`, oversize `num_lines`, zero-line read, short buffer, idempotent
`finish_decompress`, read-past-end returns `Ok(0)`)

Possible Follow-up

- Switch the streamable cases (single-SOS baseline, progressive post-process) to per-MCU-row decoding so `start_decompress` no longer materialises the full image.
- The multi-SOS interleaved baseline edge case (one component per scan) requires all components before post-processing and would stay decode-then-slice in the streaming follow-up.

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -34,6 +34,7 @@ use crate::huffman::HuffmanTable;
 use crate::idct::{choose_idct_func, choose_idct_1x1_func, choose_idct_4x4_func};
 use crate::marker::Marker;
 use crate::misc::SOFMarkers;
+use crate::scanline::ScanlineState;
 use crate::upsampler::{
     choose_horizontal_samp_function, choose_hv_samp_function, choose_v_samp_function,
     generic_sampler, upsample_no_op
@@ -232,6 +233,10 @@ pub struct JpegDecoder<T> {
     pub(crate) extended_xmp_segments: Vec<ExtendedXmpSegment>,
     /// Current decoding phase for incremental decoding.
     state: DecodingState,
+    /// Per-call state for the libjpeg-turbo-style scanline output API.
+    /// `None` until `start_decompress` is called; back to `None` after
+    /// `finish_decompress`.
+    pub(crate) scanline_state: Option<ScanlineState>
 }
 
 impl<T> JpegDecoder<T>
@@ -338,6 +343,7 @@ where
             coeff:             1,
             extended_xmp_segments: vec![],
             state:             DecodingState::DecodeHeaders { resume_position: 0 },
+            scanline_state:    None,
         }
     }
     /// Decode a buffer already in memory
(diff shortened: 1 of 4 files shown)


etemesi254:
hi

thanks for this


there is already an ongoing pr with the aim of getting that

etemesi254:
See https://github.com/etemesi254/zune-image/pull/386

ender672:
Thanks for the pointer! I'd seen #386 and tried to follow the same shape by having methods on JpegDecoder and mirroring libjpeg-turbo names. But I think these two PRs cover different libjpeg-turbo entry points.

#386 does `jpeg_read_raw_data`, and this PR does `jpeg_read_scanlines`.

My motivation is the Nautilus/glycin memory issue with large images. I want converted rows so I can scale on the fly without ever materialising the full decoded image.

etemesi254:
Hi, 

The end goal is to have a jpeg read mcu row  from the #386, as of now we are still early stages of getting it there, 

Looking at your code, it will actually make the memory usage higher, 

You are storing  the full data in 

```
   /// The fully-decoded image. Scanlines are sliced out of this on each
    /// `read_scanlines` call.
    pub(crate) pre_decoded: Vec<u8>
 ```
 So now the caller has to allocate a new buffer for which they can stream the rows from, even if this is 1 row, the total allocations for this will be (whatever zune-jpeg allocates + 1 row), this does not solve the memory problem
 
 
The libjpeg-turbo read_scanlines is a bit more work than what is here, we have to worry about a lot of things , and that is what that pr covers, there are a lot of edge cases to worry about, you need to see how that api works for progressive images, for images with the weird sampling factors, you can optimize the scanlines code by reducing per component allocations, in other words you can do a lot, this PR just adds overhead on top of the existing api without the benefits, people who are used to libjpeg-turbo will try to resuse that and be a bit shocked as to why the first row takes that long and the others are faster, and memory allocated is to store one image.

For the api you made, you can simply decode the full image, pipe it via `chunks_exact` and get the same benefits (or even more if you skip the allocation),  so there is no need to do it this way, so for now will have to not accept this PR

ender672:
Got it. You already have plans for scanline-at-a-time decoding, starting with internals. My approach here was the opposite -- start with the external API (only) and work backwards from there. Thanks for taking the time to explain!

PR #392, opened 2026-05-12
Remove newlines from error Debug implementations

### LINKED ISSUE #385: Newline in errors
The [`Debug` implementation for `ZByteIoError `](https://github.com/etemesi254/zune-image/blob/4d8dd27c20dcd464eb2a7c4805db7635915e5520/crates/zune-core/src/bytestream/reader.rs#L70) adds newlines. Was this a conscious choise? Otherwise I think it would be better if it doesn't append a newline.

If it helps, my use case: I'm instrumenting functions using the `tracing` library, like so: `#[instrument(skip_all, err)]`, and this adds empty lines in my log file. It would be worse if the error message wasn't at the end of the line, since that would split the log line.

It's easy to just strip the newline in consumer code, but I think it's generally better to avoid newlines in such formatted strings.

kimjune01:
The Debug implementations for various error types were appending newlines via `writeln!`, which splits log lines when using instrumentation libraries like `tracing`.

## Why

Error messages should not include trailing newlines - formatters like `#[instrument(skip_all, err)]` expect the error string to be inline. Newlines in the middle of a log entry make logs harder to parse and break structured logging.

## What Changed

Replace `writeln!` with `write!` in Debug implementations across all error types:
- `ZByteIoError`
- `ImageErrors` 
- `ImageDecoderErrors`
- `ChannelErrors`
- `PPMDecodeErrors`
- `PPMEncodeErrors`

Add tests to verify error messages have no trailing newlines.

Fixes #385


etemesi254:
thanks

could you remove the cargo lock on the pr

then its mergable

PR #396, opened 2026-05-18
JpegDecoder: Make marker parsing and scan replay resumable

telecos:
Follow-up to #390. Makes marker parsing atomic and scan replay idempotent so recoverable EOF mid-marker leaves the decoder retry-safe without reprocessing committed data.

Changes:

- `with_marker_body` buffers the full marker payload before mutation; EOF mid-read is a no-op on decoder state.
- All header parsers (DHT, DQT, DAC, SOF, SOS, APP, COM) stage locally then commit.
- `ScanHeaderStateSnapshot` captures first-SOS entropy/QT tables for replay rollback after inter-scan redefinitions.
- `ScanCheckpoint` extended with `SosParamsSnapshot` for exact RST-boundary resume.
- Progressive falls back to scan-start replay (refine passes make mid-scan resume unsafe).
- RST checkpoint invalidated at row boundaries and before inter-scan markers.
- `ScanCheckpoint` remains allocation-free `Copy` scalars + fixed arrays); `EntropyTables` cloned only once at scan-state entry.
- Integration tests: byte-position truncation recovery, chunk-split recovery, retry idempotency, post-success replay parity.

TO-DO in follow-Up:

- Preserve entropy decoder state within restart intervals to reduce replay work on mid-MCU EOF.
- Separately: fix pre-existing `discard_mcu_block` DC prediction loss for skipped components.

Part of #375.

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -31,10 +31,10 @@ use crate::errors::{DecodeErrors, UnsupportedSchemes};
 use crate::headers::parse_dac;
 use crate::headers::{
     parse_app1, parse_app13, parse_app14, parse_app2, parse_dqt, parse_huffman, parse_sos,
-    parse_start_of_frame
+    parse_start_of_frame, with_marker_body
 };
 use crate::huffman::HuffmanTable;
-use crate::idct::{choose_idct_func, choose_idct_1x1_func, choose_idct_4x4_func};
+use crate::idct::{choose_idct_1x1_func, choose_idct_4x4_func, choose_idct_func};
 use crate::marker::Marker;
 use crate::misc::SOFMarkers;
 use crate::upsampler::{
@@ -77,22 +77,16 @@ pub type ColorConvert16Ptr = fn(&[i16; 16], &[i16; 16], &[i16; 16], &mut [u8], &
 /// Carry out IDCT (type 3 dct) on ach block of 64 i16's
 pub type IDCTPtr = fn(&mut [i32; 64], &mut [i16], usize);
 
-/// State recorded once headers complete, used for scan-phase replay/resume.
+/// Scan-phase state kept so `decode_into` can retry or replay after SOS.
 ///
-/// Stored on the decoder behind an `Option<Box<...>>` so that one-shot
-/// decoding (the path that never resumes) carries only a null pointer's
-/// worth of overhead. The box is created on the SOS marker and either
-/// dropped when the scan completes successfully or kept alive across a
-/// recoverable EOF so the next `decode_into` call can resume.
-///
-/// `append_snapshot` rolls back inline metadata on scan replay.
-/// `sos_snapshot` restores scan parameters before replay.
-/// `rst_checkpoint` resumes from the latest restart boundary when present.
+/// Full replay starts at the first SOS; `rst_checkpoint` can resume from a
+/// later restart boundary when one is still valid.
 #[derive(Clone)]
 pub(crate) struct ScanDecodeState {
     pub(crate) scan_start_position: usize,
     pub(crate) append_snapshot:     HeaderAppendStateSnapshot,
     pub(crate) sos_snapshot:        SosParamsSnapshot,
+    pub(crate) header_snapshot:     ScanHeaderStateSnapshot,
     pub(crate) rst_checkpoint:      Option<Box<ScanCheckpoint>>
 }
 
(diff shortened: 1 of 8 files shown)


etemesi254:
Thanks, reviewing

Also while at it you add a check that ICC sections are bounded  e.g if an image says it has 50 ICC split channels thats rejected (meaning some checks on APP2)

crates/zune-jpeg/src/decoder.rs
```diff
@@ -896,26 +918,25 @@ where
...
+                // carries pixel-density metadata we currently ignore. The
+                // single thing we care about here is the Motion-JPEG marker
+                // — Microsoft's AVI/AVI2 container stores per-frame JPEGs
+                // with an `b"AVI1\0"` identifier in APP0 instead of JFIF.
+                // When we see it, we tag the decoder as MJPEG so downstream
+                // logic can apply MJPEG-specific concessions (e.g. missing
+                // DHT segments fall back to the standard tables). The
+                // body-length guard tolerates JFIF bodies shorter than five
+                // bytes by simply not matching them — anything that isn't
+                // exactly the AVI1 signature is silently skipped.
```
etemesi254, line 931: Thanks for the context (had it in my head should have put it as comment)

crates/zune-jpeg/src/decoder.rs
```diff
@@ -1255,21 +1291,45 @@ where
...
+
+        match result {
+            Ok(()) => {
+                // Drop the RST checkpoint so a post-success replay starts
+                // from scan-start with zeroed DC predictors instead of
+                // pointing at stale entropy data. `scan_state` is expected
+                // to still be `Some` here (it was set when SOS was parsed
+                // and is only cleared on hard error), but we guard with
+                // `if let` rather than `expect` so a future change that
+                // clears it on success degrades to a no-op instead of a
```
etemesi254: Is this future change expected?

crates/zune-jpeg/src/headers.rs
```diff
@@ -25,90 +25,166 @@ use crate::errors::DecodeErrors;
...
+) -> Result<R, DecodeErrors>
+where
+    T: ZByteReaderTrait,
+    F: FnOnce(&mut JpegDecoder<T>, MarkerBody<'_>) -> Result<R, DecodeErrors>
+{
+    let length = decoder.stream.get_u16_be_err()?;
+    let body_len = usize::from(length)
+        .checked_sub(2)
+        .ok_or(DecodeErrors::FormatStatic("Marker length < 2"))?;
+    let body_pos = decoder.stream.position()?;
```
etemesi254: Any reason we need the position?

The problem is that pos is a bit heavy depending on the underlying stream, on cursor its simply returning an index but if the reader is a file or something it calls seek, so now we introduce kernel context switching
 per header

crates/zune-jpeg/src/headers.rs
```diff
@@ -608,151 +640,123 @@ pub(crate) fn parse_app1<T: ZByteReaderTrait>(
...
-            .stream
-            .peek_at(EXTENDED_XMP_HEADER_SIZE, length - EXTENDED_XMP_HEADER_SIZE)?
-            .to_vec();
-
-        decoder.extended_xmp_segments.push(ExtendedXmpSegment { offset, total_size, guid, data });
-    } else {
-        warn!("Unknown format for APP1 tag, skipping");
-    }
+    enum App1Payload {
+        Exif(Vec<u8>),
```
etemesi254, line 638: Much better than what was there, appreciate it

etemesi254 (commented):
Generally well made, thanks for cleaning up the app segments, i think my only worry is `stream.position()` on the `with_marker_body`

telecos:
> Generally well made, thanks for cleaning up the app segments, i think my only worry is `stream.position()` on the `with_marker_body`

Removed `position()` from `with_marker_body` entirely. The only caller that actually needs the stream offset is the MPF path in `parse_app2`, and it now computes it lazily as `decoder.stream.position()? - body.len() as u64`. This matches libjpeg-turbo's approach in [jdmarker.c](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/src/jdmarker.c#L1067-L1084) where APP2 markers go through the generic process_APPn handler with no per-marker bookkeeping beyond what the individual handler needs.

telecos:
> Thanks, reviewing
> 
> Also while at it you add a check that ICC sections are bounded e.g if an image says it has 50 ICC split channels thats rejected (meaning some checks on APP2)

Updated to mirror libjpeg-turbo's validation flow exactly. In libjpeg-turbo, the marker parsing stage (jdmarker.c) does no ICC-specific validation — APP2 bodies are saved verbatim. All validation happens at assembly time in [jdicc.c](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/src/jdicc.c#L69-L167):

- [#define MAX_SEQ_NO 255](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/src/jdicc.c#L78) — no artificial cap, uses the full u8 range
- [if (seq_no <= 0 || seq_no > num_markers)](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/src/jdicc.c#L107) — rejects bogus sequence numbers
Inconsistent num_markers across chunks → warn + reject
- Duplicate/missing seq_no → warn + reject

We now do the same:

- Parse time (parse_app2): reject num_markers == 0 and seq_no out of [1, num_markers] — these are always invalid per-spec so failing early is safe and avoids buffering garbage. No artificial cap on num_markers.
- Assembly time (icc_profile()): checks consistency across chunks, duplicate seq_no, and missing sequences — matching libjpeg-turbo's second-pass validation.

Rationale for not capping at 50 (or any fixed value): the spec ([ICC Technical Note on Profile Embeddings 2021, JFIF section](https://archive.color.org/files/technotes/ICC-Technote-ProfileEmbedding.pdf)):

> The identifier sequence is followed by one byte indicating the sequence number of the chunk 
(counting starts at 1) and one byte indicating the total number of chunks. All chunks in the sequence 
should indicate the same total number of chunks. The 1-byte chunk count limits the size of 
embeddable profiles to 16 707 345 bytes. 

which allows up to 255 chunks, and libjpeg-turbo doesn't impose a smaller limit. We don't want behavioral divergences from libjpeg-turbo on "valid" files.

crates/zune-jpeg/src/headers.rs
telecos: Good catch, we don't need it in most cases, better to get position in the specific context we need that.

PR #397, opened 2026-05-20
JpegDecoder: Add row-granular resume for baseline Huffman scans

telecos:
Follow-up to #396. Preserves Huffman entropy decoder state at MCU-row boundaries so recoverable EOF mid-scan can resume from the last stable row instead of replaying the full scan from SOS.

Changes:

• `BitStream` now exposes save/restore hooks for entropy decoder state.
• `BitStreamHuffman` snapshots the bit buffer, aligned buffer, marker state, overread state, EOI state, and EOB run.
• `ScanCheckpoint` now carries bitstream state, restart countdown, SOS/header metadata, DC predictors, and stable output progress.
• Baseline Huffman single-SOS scans record per-row checkpoints on retry, keeping one-shot decode free of checkpoint overhead.
• RST checkpoints continue to work and now restore the restart countdown exactly.
• Vertical/HV upsampling retry state is preserved across same-size retry setup so resumed rows stay bit-identical to one-shot decode.
• `JpegDecoder::decoded_scanlines()` reports the stable output prefix for partial-display callers.
• Arithmetic streams explicitly opt out of row-granular checkpoints and keep RST/scan-start retry behavior.

Tests:

• Non-RST baseline scan truncation resumes from entropy data instead of scan start.
• Vertical/HV upsampling resume matches one-shot output.
• RST/arithmetic checkpoint tests continue to pass.
• Full incremental suite passes with `cargo test -p zune-jpeg --features arith`.

TO-DO in follow-up:

• Add an explicit streaming/incremental mode so first-attempt scan decoding can checkpoint without changing default one-shot performance.
• Complete baseline multi-SOS/non-interleaved fine-grained resume semantics.
• Add progressive JPEG partial-display and fine-grained resume support.

This work is part of #375.


etemesi254:
LGTM.

Though a question, will there be a large performance difference drift in the granular side vs the normal one shot decode?

eventually, would want to move away from one shot into the same 1 MCU row like decode, and a clear boundary with post process (I am to believe you won't be consuming this), so that I can explore multithreaded decode which has always been in my todo list

telecos:
> Though a question, will there be a large performance difference drift in the granular side vs the normal one shot decode?

I measured this specifically because the first version did regress one-shot decode.

In the initial checkpoint version (`d5030440`), where the normal decode path still paid checkpoint-related work, focused local Criterion runs showed one-shot regressions such as:

- No sampling baseline: +10.6%
- Horizontal subsampling: +6.3%
- Vertical subsampling: +7.7%
- HV subsampling: +4.8%
- Grayscale: +14.1%
- Intrinsics path: +10.2%
- DRI one-shot: +4.7%

That is why the current version treats the one-shot path specially. `mcu_checkpoints_enabled` is only enabled when `decode_into` is entered with an existing `scan_state`, i.e. on a retry/resume call after recoverable EOF. The first normal one-shot decode keeps it false and skips the per-row checkpoint work.

After that change, the same focused one-shot measurements against current `dev` are back within noise:

- No sampling baseline: -0.3%
- Horizontal subsampling: -0.7%
- Vertical subsampling: -0.5%
- HV subsampling: +0.3%
- Grayscale: -0.2%
- Intrinsics path: -0.2%
- DRI one-shot: +0.1%

So the granular resume path does have extra work when the caller is actually retrying after EOF, but normal one-shot decode should not see meaningful drift.

> eventually, would want to move away from one shot into the same 1 MCU row like decode, and a clear boundary with post process...

Agreed. I see this PR as a compatibility/resume step that avoids changing the main decode/post-process structure. A future row/MCU-boundary decode pipeline would be cleaner for streaming and would also line up better with future multithreaded decode. This PR keeps that direction open while avoiding a steady-state cost on the existing one-shot path.

etemesi254 (approved):
Redid a second pass, looks okay, no objections from me, will await your go ahead before merging in case you have anything else you may want to add

telecos:
> Redid a second pass, looks okay, no objections from me, will await your go ahead before merging in case you have anything else you may want to add

I think that we can complete this PR as is. I have already prepared some follow up on separated branch, so probably handle it in separated PR.

PR #398, opened 2026-05-22
Fix failing no_std CI

telecos:
Add missing import which is making CI to fail.


etemesi254:
Thanks, apologies was caught up in other changes so local device missed this

PR #399, opened 2026-05-26
JpegDecoder: Make incremental decode contract explicit and add checkpoint opt-in

telecos:
Follow-up to #397. This PR makes the public incremental-input contract explicit and adds an opt-in mode for callers that know they are feeding JPEG data incrementally.

The previous resumability work keeps one-shot decoding fast by default: row checkpoints are only recorded after a scan decode attempt has already happened. That means an incremental caller may pay one scan-start replay before per-row checkpoints help. This PR keeps the default one-shot path unchanged, while allowing streaming callers to opt into first-attempt row checkpoints.

Changes:

- Add `JpegDecoder::decoded_output_bytes()` to expose the stable decoded byte prefix after `decode_into()`.
- Keep `decoded_scanlines()` consistent by deriving it from `decoded_output_bytes()`.
- Document the recoverable EOF retry contract for `decode_headers()` and `decode_into()`.
- Add `JpegDecoder::set_incremental_mode(bool)` and `JpegDecoder::incremental_mode()`.
- In incremental mode, enable per-row checkpoints during the first scan decode attempt.
- Keep checkpointing conservative: baseline Huffman, non-progressive, single-SOS, supported bitstream state.
- Add tests proving:
  - recoverable EOF vs hard header failure remains distinguishable,
  - decoded byte/scanline progress is reported consistently,
  - default mode first retry replays from scan start,
  - incremental mode first retry resumes from an entropy-data row checkpoint.
- Add benchmark coverage comparing one-shot/default behavior with incremental mode and first-retry behavior.

This work is part of #375

### DIFF
--- a/benchmarks/benches/decode_jpeg.rs
+++ b/benchmarks/benches/decode_jpeg.rs
@@ -175,6 +175,13 @@ fn decode_jpeg_opts(buf: &[u8], options: DecoderOptions) -> Vec<u8> {
     d.decode().unwrap()
 }
 
+fn decode_jpeg_incremental_mode(buf: &[u8]) -> Vec<u8> {
+    let mut d = JpegDecoder::new(ZCursor::new(buf));
+    d.set_incremental_mode(true);
+
+    d.decode().unwrap()
+}
+
 fn decode_no_samp_opts(c: &mut Criterion) {
     let a = sample_path().join("test-images/jpeg/benchmarks/speed_bench.jpg");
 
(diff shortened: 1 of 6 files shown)


telecos:
@etemesi254 for some reason CI workflows did not run automatically on this PR as it used to happen

etemesi254:
Hmm, not sure wht its running, let me investigate

etemesi254:
the only change i can think of was some CI change to reduce times by running on less platforms

probably should redo it to see if I accidentally removed the pr trigger

etemesi254:
I see the changes are minimal here with just exposing the incremental decode knobs

The only thing I would suggest is we add the same example in the README to lib.rs so that docs.rs can render it too.

Will await yout approval in case there is anything you may want to run

telecos:
> same example in the README to lib.rs so that docs.rs can render it too.
> 
> Will await yout approval in case there is anything you may want to run

Thanks, good catch. I added the incremental-input example to the crate-level docs as well, so docs.rs will render the same guidance that is in the README.

I kept this PR intentionally small and focused on the public incremental-input contract: the recoverable EOF retry semantics, stable decoded-output reporting, and the explicit checkpoint opt-in. The next follow-up has broader baseline coverage/fallback behavior changes, and mixing that into this PR would make the API surface and its tradeoffs less obvious to review.

etemesi254:
Then the CI runs.

Odd

PR #400, opened 2026-06-02
JpegDecoder: Complete baseline incremental checkpoint coverage

telecos:
This PR extends baseline JPEG incremental decoding so row-granular checkpoints also work for baseline multi-SOS / non-interleaved images.

Previously, baseline row checkpoints were limited to single-SOS scans because multi-SOS scans could not safely expose stable output until all component scans had been decoded. This change keeps the stable-output contract intact while allowing scan-body resume within each baseline multi-SOS scan, using the decoder-owned coefficient buffers that persist across retries.

What Changed

- Enables baseline row checkpoints for multi-SOS / non-interleaved scan bodies.
- Preserves scan checkpoints across inter-scan marker parsing until the next SOS is fully parsed.
- Rolls back inter-scan marker side effects on recoverable EOF so retries can replay markers safely.
- Keeps default one-shot decode on the low-overhead path by only capturing rollback snapshots when an active scan checkpoint exists.
- Updates incremental decoding docs to clarify multi-SOS checkpoint behavior and stable-output semantics.
- Adds coverage for:
  - scan-body resume in non-interleaved 4:4:4, 4:2:0, 4:2:2, and 4:4:0 JPEGs
  - recoverable EOF at multi-SOS marker boundaries
  - restart-marker checkpoint resume behavior

This work is part of #375

### DIFF
--- a/crates/zune-jpeg/README.md
+++ b/crates/zune-jpeg/README.md
@@ -47,6 +47,11 @@ caller expects input to arrive incrementally; this records checkpoints during
 the first baseline Huffman scan attempt and can reduce replay work on the next
 retry.
 
+Fine-grained row checkpoints currently apply within baseline Huffman scan
+bodies, including baseline multi-SOS / non-interleaved images. Those images may
+still report no stable output rows until the later component scans have been
+decoded and final assembly has run.
+
 ```Rust
 use zune_core::bytestream::ZCursor;
 use zune_jpeg::JpegDecoder;
--- a/crates/zune-jpeg/README.md
+++ b/crates/zune-jpeg/README.md
@@ -52,6 +52,13 @@ bodies, including baseline multi-SOS / non-interleaved images. Those images may
 still report no stable output rows until the later component scans have been
 decoded and final assembly has run.
 
+Scan checkpoints store only scalar resume state: stream position, next MCU
+row/column, restart countdown, SOS parameters, DC predictors, and bitstream
+state. Coefficients for already-decoded component scans stay on the decoder
+across retries. For multi-SOS images this means a retry can continue inside the
+current component scan, then decode later component scans, but output rows are
+not considered stable until final assembly has all component data.
+
 ```Rust
 use zune_core::bytestream::ZCursor;
 use zune_jpeg::JpegDecoder;
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -79,7 +79,7 @@ pub type IDCTPtr = fn(&mut [i32; 64], &mut [i16], usize);
 
 /// Scan-phase state kept so `decode_into` can retry or replay after SOS.
 ///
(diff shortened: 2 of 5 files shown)


etemesi254:
Hi, thanks

So what is being saved per snapshot?


E.g I have a 3 multi component/seperatable SOS image.

And I am decoding component 1 and need to feed in more data (so the snapshot logic is running), on resume, do Icontinue on  decoding component 1 and then finish component 2 and 3 and then start returning pixels? (I.e we fully decoded the image so subsequent calls just chunk the data for you if its row wise)

Or is it different?

telecos:
> Hi, thanks
> 
> So what is being saved per snapshot?
> 
> E.g I have a 3 multi component/seperatable SOS image.
> 
> And I am decoding component 1 and need to feed in more data (so the snapshot logic is running), on resume, do Icontinue on decoding component 1 and then finish component 2 and 3 and then start returning pixels? (I.e we fully decoded the image so subsequent calls just chunk the data for you if its row wise)
> 
> Or is it different?

Thanks, this question exposed a naming/documentation gap in the PR.

The checkpoint is not a full coefficient snapshot. It stores only scalar resume state: stream position, next MCU row/column, restart countdown, SOS/header state, DC predictors, and bitstream state. The decoded coefficient/sample buffers remain owned by the decoder across retries.

So for a 3-component separable SOS image, if EOF happens while decoding component 1, the retry resumes inside that component scan from the latest valid scan checkpoint when available. Once component 1 completes, the decoder continues parsing/decoding component 2 and component 3. However, for this multi-SOS baseline path, `decoded_output_bytes()` / `decoded_scanlines()` still report no stable output until all needed component scans have been decoded and final assembly has run.

I pushed a follow-up cleanup commit locally to make that clearer in the code and docs:
- renamed the internal `rst_checkpoint` field to `scan_checkpoint`, since the checkpoint may be a restart or MCU-row boundary;
- updated baseline comments from RST-only terminology to scan-checkpoint terminology;
- documented that scan checkpoints save scalar state only, while coefficient buffers stay on the decoder;
- documented the multi-SOS behavior that resume can continue inside the current component scan, but stable output is not exposed until final assembly.

For the case of incremental decoding in which we would like to “show something as soon as possible,” I think we should keep this PR’s stable-output contract intact and add a separate preview/progress surface in follow-up work. That future API can expose displayable-but-replaceable preview pixels, for example progressive previews after completed scans, without changing the meaning of `decoded_output_bytes()` / `decoded_scanlines()` as final stable output.

etemesi254:
Ah, makes very much more sense.

etemesi254:
Okay, reviewed, okay, awaiting your final go ahead before merging

telecos:
> Okay, reviewed, okay, awaiting your final go ahead before merging

Thanks for the review.

From my side, this PR is ready to merge. I would keep this PR focused on baseline incremental checkpoint coverage and the stable-output contract.

In the follow-up PR I plan to work on progressive JPEG preview / partial display, with preview output kept separate from stable output.

PR #401, opened 2026-06-07
JpegDecoder: Add progressive incremental preview display

telecos:
This PR adds progressive JPEG preview display support for incremental decoding while preserving the existing stable-output contract.

Progressive scans are now decoded through a scratch coefficient buffer and committed only after a scan completes. On recoverable EOF, the decoder can render the latest fully completed progressive scan as a replaceable full-frame preview without exposing partially decoded AC/refinement data or double-applying refinement bits on retry.

This adds separate preview/progress APIs:

- `decoded_scans()`
- `decoded_preview_output_bytes()`
- `decoded_preview_scanlines()`

The existing stable-output APIs remain final-output-only:

- `decoded_output_bytes()`
- `decoded_scanlines()`

For progressive JPEGs, the stable output prefix stays zero until the full image completes. Preview pixels are exposed separately and may be replaced by later scans.

## Details

- Add progressive scan-boundary checkpoint state for incremental retry.
- Keep committed progressive coefficients on the decoder and decode the active scan into scratch storage.
- Render partial previews only from completed progressive scans.
- Preserve the last completed preview across repeated recoverable EOF retries.
- Clarify README and crate docs for stable output vs progressive preview semantics.
- Document `None` vs `Some(0)` behavior for progressive preview APIs.
- Add tests covering:
  - progressive final parity,
  - completed DC scan display,
  - AC scan retry behavior,
  - refinement retry idempotency,
  - baseline images not exposing progressive preview state.

## Scope

This PR is focused on progressive preview correctness and API separation.

It intentionally does not add fine-grained progressive resume performance optimizations inside a scan. Progressive retries restart from the current scan boundary to avoid replaying partial read-modify-write refinement state incorrectly. That optimization work is left for the follow-up progressive resume/performance PR.

This PR is part of #375


etemesi254:
apologies for delay, will review over the weekend

etemesi254:
So from a high level overview, does that mean there is an extra step on progressive decoding where we write to the preview buffer then continue?

telecos:
> So from a high level overview, does that mean there is an extra step on progressive decoding where we write to the preview buffer then continue?

Not quite as a separate preview-buffer step in the normal progressive path.

The decoder keeps committed coefficient buffers for completed scans. For the currently active progressive scan, it clones the relevant committed coefficients into scratch coefficient storage and applies the scan updates there. If the scan completes, that scratch state is committed. If input ends mid-scan, the scratch state is discarded, so partial AC/refinement updates are not exposed or applied twice on retry.

When recoverable EOF happens and at least one progressive scan has completed, the decoder may render those committed coefficients into the caller-provided decode_into output buffer as a full-frame replaceable preview. That preview is reported through the decoded_preview_* APIs only. The stable-output APIs still report zero bytes for progressive images until final decode completes.

So the extra work is the scan-boundary coefficient scratch/commit step, plus optional preview rendering from committed coefficients on EOF. There is no separate permanent decoder-owned preview pixel buffer that every scan writes through before continuing.

etemesi254:
I believe the commited coefficient buffers were there previously, these were the huge allocated ones the decoder has to do to store the coefficient per pass

> The decoder keeps committed coefficient buffers for completed scans. For the currently active progressive scan, it clones the relevant committed coefficients into scratch coefficient storage and applies the scan updates there. If the scan completes, that scratch state is committed. If input ends mid-scan, the scratch state is discarded, so partial AC/refinement updates are not exposed or applied twice on retry.

If there is cloning of huge buffers, would love it be opt in rather than always there(I see the example uses `set_incremental(true)` so i hope this covers opt in behaviour), the other users of the library are more fire and forget so don't see the need for them to pay for something they may never use (unless I am mistaken on the maximum size of the coefficient buffer)

Also a bit of clarity

The buffers are raw plane data, no upsampling and color conversion has been done?

If so should we clarify that is what we are doing?  or is there a part I may be missing, my thoughts are if they say you can do incremental and I feed one byte at a time, say every 10 ms, when i start receiving output am I to do the conversion or should I expect it has been done.

And if so, was wondering if Chromium would also consume the upsampling->Color conversion paths, from my initial understanding most of that was  to be offloaded to the GPU samplers present in chromium code

telecos:
> I believe the commited coefficient buffers were there previously, these were the huge allocated ones the decoder has to do to store the coefficient per pass

Yes, exactly. The committed coefficient buffers were already part of progressive decoding. I should have been clearer there: this PR is not introducing those main coefficient buffers. What it adds is the optional scan-preservation behavior around the active progressive scan, so a partially decoded AC/refinement scan is not committed if input ends mid-scan.

> If there is cloning of huge buffers, would love it be opt in rather than always there

Agreed. I updated the branch so the first progressive scan attempt keeps the direct coefficient path by default. The scratch/commit preservation path is now enabled on the first attempt only when the caller opts in with `set_incremental_mode(true)`.

That keeps fire-and-forget one-shot callers on the low-overhead path. If a caller has already hit recoverable EOF and retries, we follow the existing incremental recovery gate, since at that point the decoder is being used incrementally.

> The buffers are raw plane data, no upsampling and color conversion has been done?

For the internal buffers, yes, with one small wording tweak: they are raw DCT coefficient planes, so IDCT, upsampling, and color conversion have not happened there.

> when i start receiving output am I to do the conversion or should I expect it has been done.

The caller-visible preview bytes are already converted. `decoded_preview_output_bytes()` reports bytes in the same `decode_into()` output buffer, after the normal IDCT, upsampling, and color-conversion pipeline has rendered the committed coefficients into the requested output colorspace. So a caller displaying `pixels[..preview_bytes]` does not need to do conversion themselves.

> Chromium would also consume the upsampling->Color conversion paths ... most of that was to be offloaded to the GPU samplers present in chromium code

With the current `zune-jpeg` API shape, this PR exposes CPU-rendered preview pixels through the normal output buffer. A Chromium-style raw-plane/GPU-sampling path would be a different API surface, because we would need to expose coefficient or planar image data plus enough sampling/color metadata for the caller to do that work externally. I think that is worth keeping separate from this PR (probably part of #386 or a follow up of that one).

I also ran the JPEG Criterion benchmarks against a fresh `upstream/dev` baseline after this change. Criterion reported no `zune-jpeg` performance regressions; progressive one-shot paths were unchanged or within noise, and the incremental-mode paths were unchanged or within noise as well.

etemesi254:
Oh, nice 

Just one quick question, so does that mean the decode_output_bytes will always be outputting upsampled and color converted images? but now  MCU wise where supported?

telecos:
> does that mean the decode_output_bytes will always be outputting upsampled and color converted images? but now MCU wise where supported?

Yes, with one wording clarification: `decoded_output_bytes()` does not produce bytes itself; it reports how many bytes in the caller-provided `decode_into()` output buffer are stable.

Those bytes are always in the normal decoded output format, so they have already gone through IDCT, upsampling, and color conversion into the requested output colorspace. The caller does not need to do YCbCr conversion or upsampling for bytes reported through `decoded_output_bytes()`.

The “where supported” part is mostly about stability/granularity:

- For baseline images, when the decoder can prove a prefix is stable, `decoded_output_bytes()` / `decoded_scanlines()` may advance as scanlines/MCU rows are completed.
- For baseline multi-SOS/non-interleaved cases, output may still stay at `0` until final assembly, because earlier component scans are not independently display-stable.
- For progressive images in this PR, `decoded_output_bytes()` intentionally remains `0` until the full image completes, because progressive output is replaceable rather than stable.

Progressive previews are reported separately through `decoded_preview_output_bytes()` / `decoded_preview_scanlines()`. Those preview bytes are also already IDCT-processed, upsampled, and color-converted, but they represent a replaceable full-frame preview from completed scans rather than a stable appendable prefix.

For the remaining follow-up PRs, I expect the split to stay the same:

- this PR keeps progressive previews on the existing `decode_into()` pixel-output contract;
- the fine-grained progressive resume follow-up can improve where progressive retries resume, but it should not change what `decoded_output_bytes()` means;
- any raw-plane / GPU-sampling path would belong to the raw output work, probably #386 or a follow-up there, because that would expose a different output contract from the converted `decode_into()` buffer.

telecos:
Resolved conflicts. CI failure should be fixed in #415

etemesi254:
Hi, most part of the PR is LGTM on my side, just not sure about the cloning pointed above, is that always there?

telecos:
> Hi, most part of the PR is LGTM on my side, just not sure about the cloning pointed above, is that always there?

No, the coefficient cloning is not always performed.

On the default path, `incremental_mode` is disabled and the first `decode_into()` attempt decodes progressive scans directly into the decoder’s existing coefficient buffers. That preserves the current one-shot behavior and does not allocate or clone an additional full coefficient buffer.

The scratch path is used in two cases:

1. The caller explicitly enables `set_incremental_mode(true)` before the first decode attempt.
The same decoder is retried after a previous scan decode attempt, where preserving the last completed scan is necessary for correct and idempotent recovery.
1. Even on the scratch path, we do not clone every component buffer. We identify the components referenced by the current SOS and clone only those coefficient planes. If the scan completes, the scratch buffers are committed by swapping them into the decoder state. If input ends partway through the scan, they are discarded, leaving the coefficients from completed scans unchanged.

So fire-and-forget callers using the default configuration do not pay the cloning cost. The additional memory and copying are limited to incremental progressive decoding and retry preservation. I’ve clarified this distinction in the implementation and the `set_incremental_mode` documentation.

etemesi254:
Okay

that explains it


then an okay from my side

will let you rescan and then you can give the go-ahead.


Now what would be remaining from the goal of incremental decoding?

telecos:
Thanks! I think this PR is already ready to land. As a follow up, I am finish to polish a new PR based on this branch which incorporates more fine grained resumability support, but I think this has lower priority.

With #401, the core recoverable-EOF contract is largely complete:
- header and marker parsing can resume safely;
- baseline Huffman scans can resume at row/restart boundaries;
- baseline multi-SOS decoding preserves its component state;
- progressive decoding preserves completed scans and can expose them as replaceable previews;
- retries avoid committing partial progressive refinement data.

The main decoder-side item still remaining is fine-grained resume within an active progressive scan. In current PR, progressive retries resume from the current scan boundary. This is correct, but it may repeat work when EOF occurs late in a large scan. A follow-up could checkpoint progressive Huffman decoding at safe MCU-row boundaries, including the more delicate AC refinement case. Paths where restoring entropy/refinement state is not yet proven safe can retain scan-boundary replay as the correctness fallback.

After that, there is an important integration phase. I am on the process of completing integration of zune-jpeg in skia (from which we will use it in Chromium) in https://skia-review.googlesource.com/c/skia/+/1189316. The current Skia PR satisfies Skia’s incremental interface through rewind and re-decode, and explicitly lists “true incremental decoding” as requiring this upstream zune-jpeg work. A follow-up Skia CL will need to retain the same decoder and output storage across input growth, map recoverable EOF and the stable/preview progress APIs into SkCodec, and verify the behavior against Skia’s stream, rewind, cancellation, and buffer-lifetime expectations.

I expect that integration to require some iteration even after the library-side implementation is theoretically complete. Exercising the API through the Rust/C++ FFI and Skia’s real decoding lifecycle may reveal assumptions or edge cases that isolated zune-jpeg tests do not cover. We should leave room for focused upstream fixes discovered during that work.

There are also related efforts, but I see them as adjacent rather than blockers for the core recoverable-EOF goal:

- raw YUV plane output in #386, needed for Skia’s onQueryYUVAInfo/GPU conversion path;
- cooperative cancellation and progressive cancellation resume in #403/#417;
- additional integration fuzzing, performance measurements, and memory-pressure testing.

So my summary would be: after #401, incremental decoding is functionally usable and retry-safe, while efficient progressive in-scan resume and validation through the real Skia integration are the principal remaining pieces before calling the broader effort complete.

etemesi254:
Sounds very good to me

Issue #402, opened 2026-06-12
Open to a PR adding cooperative cancellation to zune-jpeg? (no new dependency)


lilith:
Would you take a PR adding cooperative cancellation to zune-jpeg's long-running entry points (`JpegDecoder::decode` / `decode_into`)?

Context: zune-jpeg sits in my image server's dependency tree, and it's also the DCTDecode path in hayro's PDF rendering (via hayro-syntax), where I'm wiring cancellation end to end (LaurenzV/hayro#1052). When a client disconnects mid-request I'd like to stop the work instead of letting it finish.

The shape I have in mind is the one from [`enough`](https://github.com/imazen/enough) — my crate, but you wouldn't depend on it: [ZERO-DEP.md](https://github.com/imazen/enough/blob/main/ZERO-DEP.md) is the vendor-one-file version. Two types, ~100 lines of code once you strip the docs, MIT/Apache-2.0. The decoder holds a stop handle and polls it every N MCU rows; on cancel it returns early with a distinct error variant. Cooperative only — no threads, no async, no unsafe.

When nobody's cancelling, the handle is `None` inside and a check is one predicted branch, so existing callers pay basically nothing. Callers that do want it can bridge from an `Arc<AtomicBool>`, tokio's `CancellationToken`, etc. in a line. Existing API stays untouched — I'd add a `decode_with_stop()`-style variant or a `DecoderOptions` setting, whatever fits your style. If even a vendored type is more than you want, a bare closure parameter (`Option<Arc<dyn Fn() -> bool + Send + Sync>>`) works too — a closure can wrap whatever cancellation source the caller already has, which a plain flag can't.

I just did the same for miniz_oxide (Frommi/miniz_oxide#193, with two ready branches — one optional-dep, one fully vendored — and measured no-regression perf). Happy to match whichever shape you prefer here. If it's out of scope, no hard feelings — close away.

PR #403, opened 2026-06-12
zune-jpeg: cooperative cancellation via a vendored StopCheck

lilith:
Follow-up to #402, as a draft — working code seemed easier to evaluate than the proposal. Feel free to close if this isn't a direction you want.

What it does:

- adds `zune-jpeg/src/stop.rs`, one self-contained type (`StopCheck`), no new dependencies, following the zero-dependency pattern from [ZERO-DEP.md](https://github.com/imazen/enough/blob/main/ZERO-DEP.md)
- `JpegDecoder::set_stop_check()` — the check is polled about every 1024 MCUs of decoding work (a countdown decremented once per MCU row in the baseline path and in each progressive scan pass), which bounds the work between polls to roughly a quarter megapixel. When it fires, decoding returns the new `DecodeErrors::Stopped`.
- the default is `StopCheck::none()`; the bookkeeping is about one subtraction and compare per MCU row, nothing per MCU or per symbol

Caveats I'm aware of:

- `DecodeErrors` isn't `#[non_exhaustive]`, so the new variant is next-0.x material — targeting `dev` for that reason.
- I couldn't run the repo's rustfmt config on a stable toolchain (unstable options), so style nits are likely; point them out and I'll fix them.
- Tests decode `medium_no_samp_2500x1786.jpg` and the limited-progressive Kiara file from `test-images`, with a counting check that fires mid-decode.

Measured on a 33 MP decode (Ryzen 9 7950X, median of 24 runs, same session): 85.3 ms on `dev` vs 86.0 and 83.1 ms across two runs of this branch — the branch brackets the baseline, so I'd only claim "below my measurement noise" rather than any particular delta. Same-binary, a live never-firing check vs none differed by under 0.5%. One machine; happy to share the harness or rerun with whatever methodology you prefer.

If you'd rather have this shaped differently (a `DecoderOptions` setting, different naming, an error shape that avoids touching `DecodeErrors`), I'm glad to rework it.

### DIFF
--- a/crates/zune-jpeg/src/cancel.rs
+++ b/crates/zune-jpeg/src/cancel.rs
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2026.
+ *
+ * This software is free software; You can redistribute it or modify it under terms of the MIT, Apache License or Zlib license
+ */
+
+//! Optional cooperative cancellation for long-running decodes.
+//!
+//! [`CancelCheck`] is a minimal, dependency-free trait the decoder polls at
+//! coarse intervals (roughly once per MCU row) to decide whether to stop. Any
+//! `Fn() -> bool` that is `Send + Sync` implements it, and [`NeverCancel`] costs
+//! nothing to poll, so the check can be threaded through internals
+//! unconditionally. Its shape is modelled on the `enough` crate's `Stop` trait
+//! (<https://github.com/imazen/enough/blob/main/ZERO-DEP.md>).
+
+use alloc::sync::Arc;
+
+/// How many MCUs are decoded between polls of the cancel check.
+///
+/// At up to 16x16 pixels per MCU this bounds the work between polls to about
+/// a quarter megapixel while keeping the check off the per-MCU path.
+pub(crate) const CANCEL_POLL_INTERVAL_MCUS: usize = 1024;
+
+/// A cooperative cancellation check, polled periodically during a long decode.
+///
+/// Set one with [`JpegDecoder::set_cancel`](crate::JpegDecoder::set_cancel);
+/// when it fires, decoding returns
+/// [`DecodeErrors::Cancelled`](crate::errors::DecodeErrors::Cancelled).
+///
+/// Any `Fn() -> bool` that is `Send + Sync` implements `CancelCheck`, so the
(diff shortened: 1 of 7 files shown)


etemesi254:
Hi,

Thanks for this.

I believe the goal is to simply stop executions via an external implementer?

lilith:
Yes

etemesi254:
Very much okay with the workings, and 1024 check makes sense, but wondering how bad of a perf regression it would be if we allowed dynamic measure (i.e not just 1024, set what you want), maybe we can benchmark on that

Also on merging hope we can wait until we get https://github.com/etemesi254/zune-image/pull/401 merged hopefully within the next one week then we can merge this

lilith:
Tbh, the 1024 is not really needed and more of a security blanket; you
could call per mcu. Implementations usually resolve to a boolean check -
but you could do whatever or make it configurable. Do you have a
preference? The enough crate has a more advanced debounce implementation.
That said, I've benchmarked generics vs callbacks and it's basically
impossible to benchmark a difference because computers are simply too good
at optimizing this away at every level.

etemesi254:
Yea compilers are pretty good at the optimizations

my goto these days has been a cheap vps there it is easier to see cache effects or branch mispreduction penalties

lilith:
Likewise, I keep some arm and x86 hertzner boxes on retainer

lilith:
I added a configurable interval debounce to #406

lilith:
Should I add some kind of progress reporting?

etemesi254:
How is progress reporting done?

Also then if so can we make this whole feature set to be opt in, via a features cargo impl so that for those who do not want it they do not have to pay the (negligible) cost of some paths?

I am thinking if progress report that means its an atomic operation, no problem with that but on hot paths I have learnt we better keep them free

lilith:
For progress reporting I think you would just add a f32 parameter and rename it to ProgressCancelToken. Progressive might require a little math. 

Compile time features have a permanent cost for testing and back-compat that isn't worth it, especially for something that is a single if statement unless used, imo.

etemesi254:
Hi, sorry it has taken a while, ready to merge this, lets fix the conflicts

lilith:
Looks like this was somehow already merged — closing.

etemesi254:
Oh yea sorry, merged the other version, thanks.

PR #405, opened 2026-06-16
isobmff: initial version refactoring from zune-heic

bradh:
As discussed at https://github.com/etemesi254/zune-image/discussions/404 this factors out some of the ISOBMFF box header parsing code from the HEIC crate.

There is still some testing and documentation required on this.


etemesi254:
thanks

one small ask

could we check what mp4 parse has snd we can borrow to make it ergonomic?

bradh:
> could we check what mp4 parse has snd we can borrow to make it ergonomic?

I'm not sure I fully understood, but I don't see a lot of difference to the equivalent in mp4parse:

https://github.com/mozilla/mp4parse-rust/blob/b693c7e4a91d1d5a391c0404d18b6fc94714894e/mp4parse/src/lib.rs#L2247

One thing I do like in the implementation is to check that the "to end" part is only used for things like `mdat`:
https://github.com/mozilla/mp4parse-rust/blob/b693c7e4a91d1d5a391c0404d18b6fc94714894e/mp4parse/src/lib.rs#L2253


They do have some convenience functions to help with the version and flags in a full box:

https://github.com/mozilla/mp4parse-rust/blob/b693c7e4a91d1d5a391c0404d18b6fc94714894e/mp4parse/src/lib.rs#L2299

and

https://github.com/mozilla/mp4parse-rust/blob/b693c7e4a91d1d5a391c0404d18b6fc94714894e/mp4parse/src/lib.rs#L2311

(I don't need those for the JUMBF implementation, but its certainly common in HEIF).

We could also do the "skip over box" and "skip rest of box" functions:

https://github.com/mozilla/mp4parse-rust/blob/b693c7e4a91d1d5a391c0404d18b6fc94714894e/mp4parse/src/lib.rs#L2322

and

https://github.com/mozilla/mp4parse-rust/blob/b693c7e4a91d1d5a391c0404d18b6fc94714894e/mp4parse/src/lib.rs#L2337

(I have something like the "skip box" implementation in my JUMBF code, and ISO/IEC 14496-12:2026 adds some words about always skipping extra bytes introduced by later versions, which seems like a terrible idea, but I'm sure MPEG knows what they're doing).

I also looked at how mp4-atom (disclosure - I'm a committer on the project, although I didn't do this part) handles it:

https://github.com/kixelated/mp4-atom/blob/8449bffae16389d35d9f3cba9387315924641c05/src/header.rs#L7

So its much simpler representation - there is no UUID handling for example. Also sizes are always `Option<u64>` and while it works well for that code base, it leans away from being a general purpose ISOBMFF reader.

bradh:
Also, can you approve the workflow ?  I'd like to know what the CI thinks of this change.

bradh:
I've added documentation and some tests. Ready for a review.

etemesi254:
Nice, thanks LGTM

Will allow you to do one pass and if you are satisfied you can reply here and I will merge

bradh:
I've made an additional change (to not have `pub` members in the `BoxHeader` structure), and adapted HEIC to match.

Should be OK to merge (or squash, as you see fit) for this round. More functionality to come in small chunks.

PR #406, opened 2026-06-18
zune-jpeg: configurable stop-check poll interval

lilith:
Stacked on #403 — adds the configurable poll interval you asked about there:

> 1024 check makes sense, but wondering how bad of a perf regression it would be if we allowed dynamic measure (i.e not just 1024, set what you want), maybe we can benchmark on that

## What

The stop-check poll interval was a hard-coded `STOP_POLL_INTERVAL_MCUS = 1024`. The decoder now exposes it, defaulting to the same 1024:

```rust
let mut dec = JpegDecoder::new(ZCursor::new(data));
dec.set_stop_check(StopCheck::from_atomic(flag));
dec.set_stop_interval(256);   // poll every ~256 MCUs (default 1024; 1 = every MCU row)
let pixels = dec.decode()?;
```

`StopCheck` itself is unchanged — the interval is decoder-side state (`set_stop_interval` / `stop_interval`) feeding the existing stack-local `Debounced` guard, not baked into the shared, `Arc`-backed cancellation handle. `0` is clamped to `1`.

## Why it's free

The interval already reached the hot loop as a runtime value — each decode loop computes `STOP_POLL_INTERVAL_MCUS / mcu_width`, and `mcu_width` is runtime — so reading it from a decoder field instead of a constant produces identical hot-loop codegen. The only difference is one field load per scan when building the `Debounced` guard.

## Benchmark

Interleaved decode of `large_no_samp_7680_4320.jpg` (33 MP baseline, 60 round-robin rounds rotating variant order to neutralise turbo bias, `--release`, no `target-cpu=native`), with a live-but-never-firing check at each interval:

| variant | median ms | min ms |
|---|---|---|
| none (no check) | 85.72 | 80.97 |
| live, interval=1 (every MCU row) | 85.92 | 81.22 |
| live, interval=16 | 85.54 | 81.37 |
| live, interval=1024 (default) | 85.94 | 81.31 |
| live, interval=MAX (poll once) | 86.04 | 81.33 |

`interval=1` vs `interval=1024`: **-0.01%** on the median — inside the ~5 ms within-variant spread. Even polling every MCU row doesn't regress, matching your read that compilers optimise this away. (Measured on a Ryzen 9 7950X; happy to rerun on a quieter VPS.)

## Tests

`crates/zune-jpeg/tests/stop.rs` adds: interval round-trip + zero-clamp, that the interval never changes decoded pixels (baseline + progressive), and that interval=1 still stops.

## Note on stacking

This is stacked on #403, so until that merges the diff here also shows the `StopCheck` commit — the only new commit is `7d5f5d28`. Once #403 lands in `dev` the extra commit drops off automatically.

### DIFF
--- a/crates/zune-jpeg/src/cancel.rs
+++ b/crates/zune-jpeg/src/cancel.rs
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2026.
+ *
+ * This software is free software; You can redistribute it or modify it under terms of the MIT, Apache License or Zlib license
+ */
+
+//! Optional cooperative cancellation for long-running decodes.
+//!
+//! [`CancelCheck`] is a minimal, dependency-free trait the decoder polls at
+//! coarse intervals (roughly once per MCU row) to decide whether to stop. Any
+//! `Fn() -> bool` that is `Send + Sync` implements it, and [`NeverCancel`] costs
+//! nothing to poll, so the check can be threaded through internals
+//! unconditionally. Its shape is modelled on the `enough` crate's `Stop` trait
+//! (<https://github.com/imazen/enough/blob/main/ZERO-DEP.md>).
+
+use alloc::sync::Arc;
+
+/// How many MCUs are decoded between polls of the cancel check.
+///
+/// At up to 16x16 pixels per MCU this bounds the work between polls to about
+/// a quarter megapixel while keeping the check off the per-MCU path.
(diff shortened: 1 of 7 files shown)


lilith:
Hmm, in this particular case it would be pretty straightforward to add a progress callback that is per-mcu-row?

etemesi254:
Hi, thanks

So will we use this PR or the other one as a final or do we merge the first one and merge this one?

lilith:
This one

PR #410, opened 2026-06-25
feat: add native support and unit test for JPEG DNL parsing

phulstaert:
## Why I made this change
I added support for parsing JPEG images that use the DNL (Define Number of Lines) marker (`0xFFDC`) because I ran into scanned images where the height in the SOF (Start of Frame) header is set to `0`. Currently, `zune-jpeg` throws a `ZeroError` on these images and fails to decode them. 

To solve this, I added a stream scanner inside `decode_headers`. If the header height is `0`, it saves the stream position, scans ahead for the DNL marker (`0xFF 0xDC`), extracts the actual height of the image, updates the dimensions, and resets the pointer. I also registered the DNL marker in the main parse loop so it is swallowed gracefully.

## How to use it
No changes are required in the client code. Zero-height images will now decode transparently:
```rust
let mut decoder = JpegDecoder::new(Cursor::new(dnl_jpeg_bytes));
decoder.decode_headers().unwrap();
let info = decoder.info().unwrap();
assert_eq!(info.height, actual_scanned_height);
let pixels = decoder.decode().unwrap();
```

## Performance & Memory Impact
- **Decodes previously failing images** that would otherwise crash.
- **0% Overhead:** Standard decoding speed and memory usage remain completely unaffected because the scan only triggers if the header height is strictly `0`.


etemesi254:
Hi, thanks for this, will review them

etemesi254:
Hu 

yes this should be done but a bit worried on the way its done.

1. Byte by byte read while may be fast on the test may be slow on real world impl where someone decides the backing storage is a hdd file

2. Reading bytes then discarding them until we hit dnl, then walking back and re re-reading the discarded bits is a lot of wasted work


A much better approach would be lets signal that we expect a dnl marker ,(self.height=0 is already a good indication) before we start parsing the entropy data we now see if we read the dnl at that point, if we did we are good, if not we error out

phulstaert:
You have a valid point; where my implementation uses in-memory and SSD sources, this is not always the case. Also it is wasteful.
I will have a look at it and optimise the code.

phulstaert:
I addressed all your remarks and in my initial tests, I got a decent speedboost (next to being "non-seekable stream"-compatible)
Please check and see if you have any other suggestions.

PR #412, opened 2026-06-25
Implement resizing/downscaling support with unit tests

phulstaert:
## Why I made this change
I added support for downscaled decoding (factors of `1`, `2`, `4`, or `8`) because generating previews, thumbnails, or lower-resolution frames of massive JPEGs can be done much faster if we downscale during decoding rather than decoding the full image and resizing afterward.

I implemented this by choosing smaller IDCT kernels inside the decoder when a downscaling factor is selected. If the downscale factor is 4, it selects the fast `4x4` IDCT kernel; if it is 8, it selects the super fast `1x1` IDCT kernel. I then updated the color conversion loop to only sample pixels on the downscaled grid (e.g. every `n`-th pixel/row), which reduces both IDCT work and color conversion latency dramatically.

## How to use it
```rust
let mut decoder = JpegDecoder::new(Cursor::new(image_bytes));
decoder.decode_headers().unwrap();

// Downscale the image by a factor of 8 (e.g., 7680x4320 -> 960x540)
decoder.set_downscale_factor(8);

let mut pixels = vec![0u8; decoder.output_buffer_size().unwrap()];
decoder.decode_into(&mut pixels).unwrap();
```

## Performance & Memory Impact
- **Speedup:**
  - Downscale Factor 2: **1.29x speedup** (`2944 ms` / run)
  - Downscale Factor 4: **1.61x speedup** (`2358 ms` / run)
  - Downscale Factor 8: **6.24x speedup** (`610 ms` / run) using fast 1x1 IDCT.
- **Memory savings:** The output buffer memory size scales down quadratically (e.g., factor 8 uses **1/64th** of the original buffer memory size).

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -293,6 +293,8 @@ pub struct JpegDecoder<T> {
     /// Number of output bytes known to be stable after the most recent
     /// `decode_into` attempt.
     pub(crate) pixels_decoded: usize,
+    pub(crate) downscale_factor: usize,
+    pub(crate) next_original_row: usize,
     /// Persistent coefficient buffers for multi-SOS baseline decoding.
     ///
     /// Owned by the decoder so contents survive a recoverable EOF and the
@@ -556,6 +558,8 @@ where
             header_resume_position: 0,
             scan_state: None,
             pixels_decoded: 0,
+            downscale_factor: 1,
+            next_original_row: 0,
             mcu_checkpoints_enabled: false,
             incremental_mode: false,
             scan_decode_attempted: false,
@@ -619,6 +623,21 @@ where
         return Some(self.info.clone());
     }
 
+    /// Set the downscaling factor (1, 2, 4, or 8).
+    pub fn set_downscale_factor(&mut self, factor: usize) {
+        self.downscale_factor = factor;
+    }
+
+    /// Return the output width of the image taking downscaling into account.
+    pub fn output_width(&self) -> usize {
+        usize::from(self.info.width) / self.downscale_factor
+    }
+
+    /// Return the output height of the image taking downscaling into account.
+    pub fn output_height(&self) -> usize {
+        usize::from(self.info.height) / self.downscale_factor
+    }
+
     /// Return the number of bytes required to hold a decoded image frame
     /// decoded using the given input transformations
     ///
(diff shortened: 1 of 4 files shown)


lilith:
Scaling after idct produces better results and can likewise be fast when streaming.

phulstaert:
Yes, but you do want to keep the resize/scale function seperate from the idct downlscale function.

If you use it, it is way faster, but the results are different and not bit for bit identical. You will see you will lose a bit of sharpness, invisible to the human eye, but this might give a slight reduced confidence in certain visual AI applications.

Often a model is trained using Python. Python by default does not use IDCT and has sharper edges. Models will tend to rely on those edges. So when you use rust in production to do inference, this might give issues.
The solution is to create your training data using the same rust 'settings' and not let python do the resizing.

etemesi254:
Just before merging, could we confirm how libjpeg-turbo implements its resizing?

phulstaert:
Hi @etemesi254,

Yes, this conceptually matches exactly how libjpeg and libjpeg-turbo implement their rapid resizing (known as IDCT scaling) and is the reason why these are for example so fast at generating thumbnails.

Instead of computing the full 8x8 IDCT, allocating a massive full-resolution memory buffer, and then spatially resizing the image, libjpeg-turbo discards the high-frequency AC coefficients and performs a smaller, partial IDCT:

- 1/2 scale: Uses a 4x4 IDCT
- 1/4 scale: Uses a 2x2 IDCT
- 1/8 scale: Uses a 1x1 IDCT

_In this PR, I am leveraging the existing idct_4x4_func and idct_1x1_func provided by zune-jpeg to achieve this exact same optimization. (Note: For a factor of 4, since a dedicated 2x2 IDCT kernel doesn't seem to exist in the codebase yet, I use the 4x4 IDCT and skip the remaining pixels during the color conversion loop. This still yields massive memory savings and speedups)._

So yes, this brings the standard libjpeg-turbo thumbnailing logic over to this crate. Let me know if you want me to clarify or adjust anything before merging!

etemesi254:
Lets have idct2x2  if we are to do 1/4 scale

lilith:
I've found the perf benefit vanishes if you compare to doing the full idct and just scaling the ycbcr or rgb pixels blockwise, and you get a higher quality result.

crates/zune-jpeg/src/decoder.rs
```diff
@@ -619,6 +623,21 @@ where
         return Some(self.info.clone());
     }
 
+    /// Set the downscaling factor (1, 2, 4, or 8).
+    pub fn set_downscale_factor(&mut self, factor: usize) {
```
etemesi254: On this, lets do a `Result<(),Err>` with only accepting 1,2,4 or 8


Or maybe use an enum, for all possible factors

etemesi254:
> I've found the perf benefit vanishes if you compare to doing the full idct and just scaling the ycbcr or rgb pixels blockwise, and you get a higher quality result.

So the suggestion is full idct->upsample->do the scaling ?

lilith:
Yes, basically, but stream as you go - it all depends on cache locality.

etemesi254:
Oh, yea, also @phulstaert  lets add benchmarks

PR #414, opened 2026-07-17
Add OpenSlide-oriented JPEG decode controls

mahogny:
This edit enables lower-level JPEG access that is needed by https://github.com/henriksson-lab/openslide-pure-rs.
This library is able to read tissue scanner images where one frequently need to be able to decode just part of a much larger image to get performance

Edit made with Codex, then cleaned up and manually inspected. But I confess to be no expert on JPEG at all so my audit is only for code smells; function was only verified more extensively by testing vs original C++ openslide code output, on some large tissue scanner datasets

- split JPEG assembly helper
- input colorspace override
- 1/2, 1/4, and 1/8 JPEG output scaling
- rectangular region decode API
- MCU-boundary-safe upsampling for segmented rows
- example and tests covering the new behavior

This works:
- cargo check -p zune-jpeg --examples
- cargo test -p zune-jpeg worker
- cargo test -p zune-jpeg --test openslide_api

Note:
I got a zune-bmp UEFI CI failure but it seems unrelated to this edit

### DIFF
--- a/crates/zune-core/src/options.rs
+++ b/crates/zune-core/src/options.rs
@@ -6,7 +6,7 @@
 //! All supported options are put into one _Options to allow for global configurations
 //! options e.g the same  `DecoderOption` can be reused for all other decoders
 //!
-pub use decoder::DecoderOptions;
+pub use decoder::{DecoderOptions, InputColorspaceOverride, JpegScale};
 pub use encoder::EncoderOptions;
 pub use encoder::PngCompression;
 mod decoder;
--- a/crates/zune-core/src/options/decoder.rs
+++ b/crates/zune-core/src/options/decoder.rs
@@ -12,6 +12,36 @@
 use crate::bit_depth::ByteEndian;
 use crate::colorspace::ColorSpace;
 
+/// JPEG input colorspace handling.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum InputColorspaceOverride {
+    /// Infer the JPEG input colorspace from markers and component layout.
+    Auto,
+    /// Force the JPEG input colorspace after marker parsing.
+    Force(ColorSpace)
+}
+
+/// JPEG output scaling.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum JpegScale {
+    Full,
+    Half,
+    Quarter,
+    Eighth
+}
+
+impl JpegScale {
+    /// Return the scale denominator.
+    pub const fn denominator(self) -> usize {
+        match self {
+            Self::Full => 1,
+            Self::Half => 2,
+            Self::Quarter => 4,
+            Self::Eighth => 8
+        }
+    }
+}
+
 /// A decoder that can handle errors
 fn decoder_error_tolerance_mode() -> DecoderFlags {
     // similar to fast options currently, so no need to write a new one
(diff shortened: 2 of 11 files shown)


etemesi254:
Hello, thank you for this, and appreciate the effort to get this.

1. Is this similar to https://github.com/etemesi254/zune-image/pull/411 in your opinion?

mahogny:
Yes! https://github.com/etemesi254/zune-image/pull/411 is a subset of this PR. Happy to divide things up differently depending on what API you prefer!

/J

etemesi254:
What are the differences?

mahogny:
https://github.com/etemesi254/zune-image/pull/411 implements:

* rectangular region decode API

But it's missing:

* split JPEG assembly helper (was there before - now just made public. puts header + data together, with some stitching)
* input colorspace override (apparently I have some messed up JPEGs that beg for an override of what colorspace is used)
* 1/2, 1/4, and 1/8 JPEG output scaling (efficient subsampling for lower resolution reconstruction)
* MCU-boundary-safe upsampling for segmented rows (specifically, this is a fast path for upsampling when aligned to MCUs)

This we can consider dropping from my patch:

* example and tests covering the new behavior

etemesi254:
Thanks for this,

I think  ` 1/2, 1/4, and 1/8 JPEG output scaling (efficient subsampling for lower resolution reconstruction)` is implemented in a different PR (same author) https://github.com/etemesi254/zune-image/pull/412

Will evaluate both and respond in a few days.

Thank you

mahogny:
Definitely some overlap. I had an agent weed through commits since my fork, to see if there is more. Below is a commented response but in short, seems there is still some more work to push. Let me know what you wish to keep and I'll prepare a targeted PR

===

  | Feature in #414 | Already covered elsewhere? |
  |---|---|
  | Rectangular region decode | Partially — #411 (`81b79669`) |
  | 1/2, 1/4, 1/8 output scaling | Partially — #412 (`74415fc5`) |
  | Input colorspace override | No — not present in any branch or PR |
  | Split header/data JPEG assembly helper | No — not present in any branch or PR |
  | MCU-boundary-safe segmented upsampling | No — not present in any branch or PR |
  | Reduced-output ("scaled") IDCT kernels | No — not present in any branch or PR |

  So #411 and #412 are the only two commits anywhere that overlap this PR, and the overlap
  is partial in both cases. Details:

  **#411 vs. the region API here.** #411 adds a stateful `set_cropping_region(&mut self, x, y, w, h)`
  plus a `use_cropping` flag, skips IDCT for blocks that don't intersect the crop (±16px), and
  crops per row inside the color-convert closure using a `vec![0u8; ..]` temp row allocated once
  per scanline. It also changes the meaning of `output_buffer_size()` globally once a crop is set. (**I might have some dislike for a stateful API since it should pull in a &mut - it would affect multithreading ability**)

  This PR instead exposes `decode_region`/`decode_region_into`/`region_output_buffer_size` with
  explicit bounds validation, a caller-owned output buffer (no forced allocation), an explicit
  `RegionDecodeMode` so the progressive fallback is part of the contract rather than implicit,
  and a dedicated baseline path (`decode_mcu_ycbcr_baseline_region` + stripe copy) so full-image
  decode semantics are untouched.

  **#412 vs. the scaling here.** #412 reuses the existing sparse-coefficient fast paths as if they
  were scaled-output IDCTs (`len <= 1 || downscale_factor >= 8` → `idct_1x1_func`), then decimates
  rows/columns during color conversion. This PR adds actual reduced-output kernels
  (`idct_int_scaled_4x4` / `_2x2` / `_1x1` plus `choose_scaled_idct_func`), with a test asserting
  they match sampling the full scalar IDCT.

  One concrete difference that matters for my use case: #412 computes
  `output_width() = info.width / downscale_factor`, which truncates, whereas libjpeg uses the
  ceiling — `(width * scale_num + scale_denom - 1) / scale_denom`. For a 1023px-wide image at 1/8,
  #412 yields 127 where libjpeg yields 128. This PR uses `div_ceil`, so scaled tile dimensions match
  libjpeg exactly, which is what OpenSlide's tiling needs.  (**there might be a way to work around this downstream but I have to check**)

Issue #94, opened 2023-03-06
zune-inflate: possibly too strict limit check


johannesvollmer:
hi!

we observed what might be a bug in zune-inflate (if not, we would still appreciate your input over at https://github.com/johannesvollmer/exrs/issues/202)

the problem is:
setting the decoder limit to the correct byte size results in an error, even though the decoder returns that exact number of bytes when no limit is set.

specifically, we observed the following test to fail:

```rust
#[test]
pub fn exact_limit_decode_zlib(){
    let compressed = std::fs::read("zlib-compressed-binary 393216 bytes.bin").unwrap();
    let options = zune_inflate::DeflateOptions::default().set_limit(393216);
    let mut decoder = zune_inflate::DeflateDecoder::new_with_options(& compressed, options);
    let mut decompressed = decoder.decode_zlib().unwrap(); // panics when a limit is set
    dbg!(decompressed.len()); // prints 393216 when no limit is set
}
```

with this file: [zlib-compressed-binary 393216 bytes.zip](https://github.com/etemesi254/zune-image/files/10901156/zlib-compressed-binary.393216.bytes.zip)
observed in zune-inflate v0.2.3 and v0.2.50

note: decompressed without limit, it results in the following bytes:
[zlib-compressed-binary 393216 bytes decompressed.zip](https://github.com/etemesi254/zune-image/files/10901181/zlib-compressed-binary.393216.bytes.decompressed.zip)

please correct me if i made any mistake :)

Shnatsel:
Thanks for the quick fix! I'll adjust the roundtrip fuzz target to look for these kinds of issues as well.

A new release needs to be published to crates.io for `exrs` to receive the fix.

etemesi254:
Yes, was an error, and fixed.

I was using the length instead of how many bytes were written in the array, sorry for that.

The reason it was so elusive is that the input stream has extra bytes so it was always in the hot loop, the hot loop had that bug, the cold loop did not exhibit it and hence it never got hit in the testing

johannesvollmer:
great to see you reacting so quickly! thanks

etemesi254:
Made 0.2.51 for this, if semver is anything to go by, the fix should happen transparently

Shnatsel:
A quick fuzzing run didn't show any more issues like these.

PR #416, opened 2026-07-19
fix(inflate): don't reject valid stored blocks that grow past size_hint

gaoflow:
The one-shot `DeflateDecoder` (`decode_deflate` / `decode_zlib` / `decode_gzip`) rejects valid,
in-spec streams whenever an uncompressed/stored block (BTYPE=00) grows the output past the current
buffer allocation. The stored-block branch conflates "buffer too small, must grow" with "user output
limit exceeded":

```rust
if dest_offset + len + FASTCOPY_BYTES > out_block.len() {
    if dest_offset.wrapping_add(len) > out_block.len() {
        return Err(/* OutputLimitExceeded(self.options.limit, out_block.len()) */);
    }
    // otherwise resize
}
```

`out_block` starts as `vec![0; size_hint]` with the default `size_hint` of 37000, while the default
`limit` is `1 << 30`. So any level-0 (or otherwise incompressible) stream over ~37 KB from a standard
encoder aborts with `OutputLimitExceeded` — reporting the buffer size, 37000 — even though the real
output is far below the limit. The intended action on the fall-through is simply to resize.

This is a regression of the class fixed in #94 ("I was using the length instead of how many bytes were
written"), reintroduced by 0c0f87b9 ("Check for output limit size before re-allocation not after").
That commit's intent — enforce the limit before over-allocating, as a decompression-bomb guard — is
right, but it checks against `out_block.len()` rather than `self.options.limit`.

The fix checks the projected output size (`dest_offset + len`) against the real limit before growing,
then resizes unconditionally, which is what 0c0f87b9 was after. The compressed-block paths already do
exactly this (`decoder.rs` resize at ~1069, limit check at ~985/~1106) and are untouched; the streaming
decoder handles stored blocks correctly against the real budget and is untouched.

Verified with a differential harness against C zlib (flate2, zlib backend) over alphabets × sizes ×
levels 0..=9 for zlib/gzip/raw deflate: 402 spurious rejects before, 0 after, with zero byte
mismatches (the decoder only ever over-rejected — when it accepts, the bytes are correct).

Tests exercise the one-shot path across all three formats over the 37000-byte growth boundary
(including multi-block inputs), and assert the bomb guard still fires on a genuinely over-limit stream,
now reporting the projected output size instead of the buffer length.

### DIFF
--- a/crates/zune-inflate/src/decoder.rs
+++ b/crates/zune-inflate/src/decoder.rs
@@ -615,18 +615,22 @@ impl<'a> DeflateDecoder<'a> {
 
                 let start = self.stream.get_position() + self.position + self.stream.over_read;
 
-                // ensure there is enough space for a fast copy
-                if dest_offset + len + FASTCOPY_BYTES > out_block.len() {
-                    if dest_offset.wrapping_add(len) > out_block.len() {
-                        return Err(InflateDecodeErrors::new(
-                            DecodeErrorStatus::OutputLimitExceeded(
-                                self.options.limit,
-                                out_block.len(),
-                            ),
-                            out_block,
-                        ));
-                    }
-                    // and if there is not, resize
+                // Enforce the limit against the projected output size, not the buffer
+                // length — growing to fit a stored block is not a limit violation (#94).
+                let projected = dest_offset.saturating_add(len);
+
+                if projected > self.options.limit {
+                    out_block.truncate(dest_offset);
+
+                    let err_msg =
+                        DecodeErrorStatus::OutputLimitExceeded(self.options.limit, projected);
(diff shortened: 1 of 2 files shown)


etemesi254:
Hi, is the set limit respected?

binggao1230:
Yes — the limit check is now against the projected output size: `projected = dest_offset + len` of the stored block, and anything over `options.limit` returns `OutputLimitExceeded` before the buffer is resized. The old code compared against the current buffer length (`out_block.len()`), which is why a valid ~37 KB+ stream tripped it even though it was far below the limit.

etemesi254:
Thank you for this

PR #420, opened 2026-08-04
JpegDecoder: Fine-grained progressive decoding support

telecos:
This follow-up to #401 improves progressive JPEG retry performance where mid-scan resume is provably safe, measures total work across repeated retries, and makes cooperative cancellation safe to retry.

PR #401 established progressive scan commit/rollback semantics and separate preview progress reporting. This PR builds on that model by adding fine-grained checkpoints for Huffman first-DC scans, whose coefficients are assigned once
rather than modified through successive refinement.

Unsafe progressive scan types continue to replay from their scan boundary.

## Safety Model

Fine-grained checkpoints are enabled only when all of the following hold:

- incremental checkpointing is active;
- the entropy decoder supports state snapshots;
- the scan is DC-only (`Ss = 0`, `Se = 0`);
- the scan is a first pass (`Ah = 0`).

This currently limits fine-grained resume to progressive Huffman first-DC scans.

The following paths deliberately retain scan-boundary replay:

- AC first scans;
- DC refinement scans;
- AC refinement scans;
- arithmetic-coded progressive scans.

These scans may update existing coefficients or entropy contexts in ways that cannot be safely resumed from an arbitrary partial position.

## Checkpoint State

A progressive fine checkpoint stores the state required to resume decoding:

- input stream position;
- active SOS parameters and component ordering;
- completed and displayed progressive scan counters;
- next MCU row and column;
- restart countdown;
- per-component DC predictors and differences;
- Huffman bitstream state, including buffered bits, marker state, overread state, EOI state, and EOB run.

The active scan remains in decoder-owned scratch coefficient storage. Coefficients are committed to the completed-scan buffer only after the scan finishes successfully.

## Checkpoint Granularity

Eligible first-DC scans checkpoint every eight completed MCU rows.

This bounds normal replay to at most eight rows while avoiding a stream-position query and bitstream snapshot after every MCU.

Scratch coefficient storage is retained while it is needed by a recoverable EOF or cancellation retry. It is released after successful decoding or when the associated resume state is no longer usable.

## Retry Behavior

On recoverable EOF:

- a safe first-DC Huffman scan may resume from its latest fine checkpoint;
- unsafe scans resume from their scan-boundary checkpoint;
- completed scans and the latest completely rendered preview remain committed;
- partial refinement or AC data is never committed.

If another EOF occurs while retrying from a fine checkpoint, the fine checkpoint is discarded and the following attempt falls back to the retained scan-boundary checkpoint. This keeps repeated retries idempotent.

## Cooperative Cancellation

Cancellation is independent of JPEG conformance handling:

- `DecodeErrors::Cancelled` is always propagated, including in non-strict mode;
- cancellation is not classified as recoverable EOF;
- callers may replace or clear the cancellation check and retry with the same decoder and output buffer;
- stable baseline rows remain valid across cancellation;
- safe progressive fine checkpoints retain their matching scratch coefficients;
- unsafe partial progressive work replays from its scan boundary;
- direct progressive decoding discards partial in-place state before retry;
- baseline multi-SOS final assembly is cancellation-aware;
- an interrupted progressive render is not advertised as a valid preview;
- retry rerenders the frame before preview progress is published again.

This includes cancellation during entropy decoding, final assembly, progressive preview replacement, and final rendering.

## Preview and Output Contracts

The APIs introduced by #401 retain their existing meaning:

- `decoded_output_bytes()` and `decoded_scanlines()` report stable final output;
- `decoded_preview_output_bytes()` and `decoded_preview_scanlines()` report a
  replaceable progressive preview;
- `decoded_scans()` reports completed progressive scans represented by the
  current preview.

Fine-grained resume changes where entropy decoding continues, not which pixels are stable or displayable.

A progressive preview is published only after a full-frame render completes. If cancellation interrupts rendering, preview byte and scanline queries return zero until retry completes the render.


etemesi254:
Hi, thank you for this.

How does libjpeg-turbo handle it? In regards to progressive fine grained?

telecos:
> Hi, thank you for this.
> 
> How does libjpeg-turbo handle it? In regards to progressive fine grained?

Thank you. My understanding is that libjpeg-turbo handles incremental input through its classic API's suspension mechanism. A custom source manager returns `FALSE` when more data is needed, and the caller supplies more input and repeats the same API call.

For progressive Huffman decoding, libjpeg-turbo can resume at the current MCU for DC, AC, and refinement scans. It keeps entropy state transactional until the MCU completes. AC refinement requires additional logic to undo newly non-zero coefficients when suspension occurs.

Progressive display is handled separately through buffered-image mode. libjpeg-turbo stores full-image coefficient buffers and lets the application consume input and render successive scans using `jpeg_consume_input()`, `jpeg_start_output()`, and `jpeg_finish_output()`.

There are two relevant limitations:

- libjpeg-turbo's arithmetic entropy decoder explicitly does not support suspension.
- libjpeg-turbo does not provide retryable cooperative cancellation. `jpeg_abort_decompress()` resets the operation and releases the current image state.

This PR is currently more conservative than libjpeg-turbo's Huffman suspension:

- Huffman first-DC scans checkpoint every eight completed MCU rows because they only assign coefficients and can be resumed safely.
- AC and refinement scans replay from the current scan boundary to avoid applying partial updates twice.
- Arithmetic progressive scans also replay from the scan boundary.
- Cancellation preserves safe checkpoints and allows retrying the same image after clearing or replacing the cancellation check.

So this is not yet full parity with libjpeg-turbo's MCU-level Huffman suspension. Reaching that would require making each progressive Huffman MCU decode transactional, including explicit undo/idempotency handling for AC refinement. This PR introduces bounded fine-grained resume only where the current decoder can prove it safe.

telecos:
I’ve prepared a follow-up branch, https://github.com/telecos/zune-image/tree/feat/jpeg-progressive-mcu-resume, stacked on #420. It closes the remaining gap with libjpeg-turbo’s fine-grained resumability for progressive Huffman JPEGs by supporting suspension and retry at MCU boundaries across DC-first, DC-refinement, AC-first, and AC-refinement scans. Arithmetic scans retain scan-boundary replay, consistent with libjpeg-turbo.

telecos:
Some update on the Chrome integration progres:

- Skia CL already landed: https://skia-review.googlesource.com/c/skia/+/1189316
- Working on Chromium CL for wiring it (patching zune-jpeg and zune-core to include up to PR #401 changes) https://chromium-review.googlesource.com/c/chromium/src/+/7681136

Once Chromium CL lands, there will be available a feature flag in `chrome://flags` / `edge://flags` to enable use of rust jpeg (zune-jpeg for decoding and jpeg-encoder for encoding).

Experiment flight on users will not yet be started until we finalize the remaining gaps.

etemesi254:
Reviewing this, as usual will ask for your blessing before merge, you can give the goahead and if i have no issues we can merge

telecos:
> Reviewing this, as usual will ask for your blessing before merge, you can give the goahead and if i have no issues we can merge

Thanks! I validated PR on Chromium side and it passed all my validations. You can go-ahead with the merge from my side.

Issue #419, opened 2026-08-03
[zune-jpeg] Buffer overflow and bitstream starvation (striped green tint) when decoding non-interleaved multi-scan JPEGs


nohle:
### Input Image
<img width="640" height="640" alt="Input Image" src="https://github.com/user-attachments/assets/661a83a0-de9c-4d0b-84c4-a5305d3e9483" />

### Output Image Without Fix (Scenario 1)
<img width="640" height="640" alt="Image" src="https://github.com/user-attachments/assets/03164bef-fe72-4606-af55-2d96e5197fa8" />

---

## Issue Description

When decoding non-interleaved multi-scan (multi-SOS) baseline JPEGs where image components (`Y`, `Cb`, `Cr`) are stored in separate sequential scans rather than interleaved, `zune-jpeg` produces a severe **striped green tint across the entire decoded image**.

Specifically, **Scan 2 (`Cb`) over-reads its Data Unit rows and consumes the bitstream belonging to Scan 3 (`Cr`)**. Because Scan 3 is starved of bitstream data, `Cr` coefficients remain at `0` across the entire image while `Cb` is corrupted by decoding `Cr` entropy data.

Additionally, when Data Unit dimensions are corrected, for non-interleaved components with a vertical sampling factor $> 1$ (`Y` in 4:2:0), `zune-jpeg` aborts at the 50% image height mark with:

```rust
DecodeErrors::FormatStatic("Would panic on slice iteration")
```

---

## Root Cause Analysis

In `crates/zune-jpeg/src/mcu.rs`:

1. **Incorrect `scan_du_height` / `scan_du_width` (Bitstream Starvation / Striped Green Tint)**:  
   For non-interleaved scans (`all_components_in_first_scan == false`), `scan_du_height` defaulted to the global `mcu_height` (which uses `v_max`) instead of the component's actual Data Unit (DU) row height, and `scan_du_width` defaulted to `mcu_width` instead of the component's actual DU row width:
   ```rust
   // Needed calculation:
   // scan_du_height = ceil((info.height * comp.vertical_sample) / (v_max * 8))
   // scan_du_width  = ceil((info.width  * comp.horizontal_sample) / (h_max * 8))
   ```
[...]

PR #421, opened 2026-08-04
[zune-jpeg] Fix baseline non-interleaved decoding

nohle:
Addresses https://github.com/etemesi254/zune-image/issues/419.


crates/zune-jpeg/tests/non_interleaved.rs
```diff
@@ -161,3 +161,53 @@ fn decode_non_interleaved_422_65x65() {
         "Center pixel wrong: expected ~(96,96,128), got ({r},{g},{b})"
     );
 }
+
+// Test for https://github.com/etemesi254/zune-image/issues/419.
+#[test]
+fn decode_baseline_non_interleaved_420_color_balance() {
+    // Path to a 4:2:0 baseline non-interleaved multi-scan JPEG (SOF0)
+    let test_data = include_bytes!("../../../test-images/jpeg/baseline_non_interleaved.jpg");
```
telecos, line 169: Could we also add this fixture to `baseline_non_interleaved_incremental_parity()`? The fix changes row coordinates used by incremental checkpoints. I tested locally this fixture with 97-byte chunks and same-decoder retries pass across all three scans.

telecos (approved):
LGTM. Thanks for the fix!

PR #424, opened 2026-08-07
fix(core): make no-op log macros expand to a valid expression

### LINKED ISSUE #423: 0.5.15 fails to compile
Still investigating, I dont even know this project.

See my CI https://github.com/marc2332/freya/actions/runs/31202803160/job/92946584851?pr=2087#step:8:69

```rs
error: macro expansion ends with an incomplete expression: expected expression
Error:    --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/zune-jpeg-0.5.15/src/mcu_prog.rs:463:17
    |
463 |                 warn!("RST marker was not found, where expected, image may be garbled")
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected expression
    |
    = note: the macro call doesn't expand to an expression, but it can expand to a statement
help: add `;` to interpret the expansion as a statement
    |
463 |                 warn!("RST marker was not found, where expected, image may be garbled");

```

canmi21:
Fixes #423.

## Problem

The no-op logging macros in `zune-core` use an empty transcriber:

```rust
macro_rules! __warn {
    ($($arg:tt)+) => {};
}
```

The braces there are the macro_rules! delimiter, not the expansion body, so
the macro expands to nothing rather than to an empty block. Expanding to
nothing is still valid in statement position, which is why zune-core compiles
standalone and cargo publish's verify build passes — the breakage only appears
in dependents that call the macro as a trailing expression:

```log
error: macro expansion ends with an incomplete expression: expected expression
  --> zune-jpeg-0.5.15/src/mcu_prog.rs:463:17
```

zune-jpeg depends on zune-core = "0.5.1", so publishing 0.5.2 pulled the
change into every existing 0.5.15 build automatically. This currently breaks
downstream consumers including egui and slint.

All five macros (__error, __warn, __info, __debug, __trace) are
affected, in both the std and not(std) branches, so no feature combination
avoids it.

Fix

Expand to a block instead:

```rust
macro_rules! __warn {
    ($($arg:tt)+) => {{
        let _ = ::core::format_args!($($arg)+);
    }};
}
```

{{ .. }} gives a real block expression of type (), valid in both expression
and statement position. Using let _ = format_args!(..) rather than a bare
{{}} additionally keeps the arguments type-checked and marked as used, so
downstream values referenced only inside a log call don't trigger
unused_variables. This matches what the log crate does for its disabled
macros, and works under no_std.

Tests

Added regression tests in log.rs covering expression position (let _: () = { warn!("..") };) and statement position.

Verified:
- cargo test -p zune-core --lib — passes
- cargo build -p zune-core --no-default-features — passes (no_std)
- cargo check --workspace — clean
- Unmodified zune-jpeg 0.5.15 from crates.io built against this branch via
[patch.crates-io] — compiles, confirming the fix resolves #423 without any
change to zune-jpeg

Note on release

Merging this doesn't unbreak users on crates.io by itself, since they resolve
zune-core from the registry. A zune-core 0.5.3 release (and/or yanking
0.5.2) is needed for the fix to actually reach them.

Hope this will help.

### DIFF
--- a/crates/zune-core/src/log.rs
+++ b/crates/zune-core/src/log.rs
@@ -41,14 +41,24 @@ macro_rules! __log_enabled {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! __error {
-    ($($arg:tt)+) => {};
+    ($($arg:tt)+) => {{
+        // Expand to a block so the macro is usable in expression position.
+        // `{}` alone expands to *nothing*, which is a hard error for callers
+        // that use the macro as a trailing expression.
+        let _ = ::core::format_args!($($arg)+);
+    }};
 }
 
 #[cfg(not(feature = "std"))]
 #[doc(hidden)]
 #[macro_export]
 macro_rules! __error {
-    ($($arg:tt)+) => {};
+    ($($arg:tt)+) => {{
+        // Expand to a block so the macro is usable in expression position.
+        // `{}` alone expands to *nothing*, which is a hard error for callers
+        // that use the macro as a trailing expression.
+        let _ = ::core::format_args!($($arg)+);
+    }};
 }
 
 //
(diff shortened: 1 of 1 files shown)


etemesi254:
Sorry for this, 

Will do better, 
yanked the bad version

Issue #425, opened 2026-08-12
Investigate render differences w.r..t. libjpeg-turbo rendering in Chromium


telecos:
While integrating zune-jpeg in Chromium, I added a comparison test to validated that JPEG test suite provides close enough rendering result to libjpeg-turbo. This umbrella bug is aimed to track all the investigation and potential fixes work on these discrepancies.

Current discrepancies are tracked in expected resuls from `JPEGDecoderComparisonTest`:

https://chromium-review.googlesource.com/c/chromium/src/+/7681136/26/third_party/blink/renderer/platform/image-decoders/jpeg/jpeg_image_decoder_test.cc#L468

etemesi254:
Nice,

Note, we do support arithmetic jpegs, see https://github.com/etemesi254/zune-image/pull/373

Contributed by [mstoeckl](https://github.com/mstoeckl)

<img width="662" height="180" alt="Image" src="https://github.com/user-attachments/assets/e9ed0c91-9c9e-4da2-a42e-d5914f525271" />

etemesi254:
My main worry would be sampling factors, there are too many and too weird, I do not know which one is supported by libjpeg-turbo which one are outright faliures, my suggestion would be we write a script to generate all accepted and test with libjpeg-turbo as reference to avoid such things https://github.com/etemesi254/zune-image/issues/422

telecos:
> Nice,
> 
> Note, we do support arithmetic jpegs, see [#373](https://github.com/etemesi254/zune-image/pull/373)
> 
> Contributed by [mstoeckl](https://github.com/mstoeckl)
> 
[...]

PR #427, opened 2026-08-12
Match libjpeg-turbo progressive arithmetic AC scan handling

telecos:
## Summary

Fix progressive arithmetic JPEG decoding when AC coefficients are delivered across many disjoint scans, including one coefficient per scan in ascending or descending spectral order.

This closes a compatibility gap found when Chromium compared zune-jpeg with arithmetic-enabled libjpeg-turbo 3.1.0.

## Chromium compatibility gap

Chromium enabled zune-jpeg's optional `arith` feature and compared decoded output against libjpeg-turbo built with arithmetic decoding enabled.

Most valid arithmetic JPEGs differed by at most 1-2 channel values. Two valid progressive arithmetic JPEG Suite fixtures instead produced severe differences:

- `progressive_arithmetic/32x32x8_grayscale_spectral_all.jpg`
  - One DC scan followed by 63 single-coefficient AC scans in ascending order.
- `progressive_arithmetic/32x32x8_grayscale_spectral_all_reverse.jpg`
  - The same AC scans in descending order.

Before this fix, the maximum channel deltas were 233 and 219, with average deltas around 45-49.

## Root cause

Zune and libjpeg-turbo make the same arithmetic decisions while decoding these scans. The difference was how they handled scan-local arithmetic exhaustion.

When an AC-first `S0=0` decision advanced beyond the scan's `Se` value:

- libjpeg-turbo followed its nonfatal `JWRN_ARITH_BAD_CODE` path, retained the coefficients decoded so far, made the remaining MCU calls for that scan no-ops, and reinitialized arithmetic state at the next SOS.
- zune-jpeg returned `ArithmeticDecode`, which stopped the whole progressive sequence. Strict mode failed, while non-strict mode rendered an image from only the scans completed before that point.

The first affected transitions were:

- Ascending order: `Ss=Se=2`
- Descending order: `Ss=Se=59`

## Fix

`decode_mcu_ac_first()` now returns its existing scan-termination signal for this condition instead of escalating it into a decoder error.

The progressive scan loop consumes that signal by ending only the active scan. Previously decoded coefficients are committed, remaining MCUs in that scan are skipped, and arithmetic registers/statistics are reset normally at the next SOS.

The Huffman path and the public API are unchanged.

The patch also imports `alloc::string::ToString` so the optional arithmetic implementation continues compiling in no-std configurations.

## Reference corpus

The tests import CC0-1.0 files from `robert-ancell/jpegsuite` revision:

`8382e7831896cd1adf1cc61a1c9e565c4030aa43`

These files were verified byte-for-byte against Chromium's mirror under:

`third_party/blink/web_tests/images/jpeg-suite/`

Primary regressions:

- `progressive_arithmetic/32x32x8_grayscale_spectral_all.jpg`
- `progressive_arithmetic/32x32x8_grayscale_spectral_all_reverse.jpg`

Valid arithmetic controls:

- `32x32x8_grayscale_spectral.jpg`
- `32x32x8_grayscale_successive.jpg`
- `32x32x8_grayscale_successive_ac.jpg`
- `32x32x8_grayscale_successive_dc.jpg`

Huffman controls:

- `progressive_huffman/32x32x8_grayscale_spectral_all.jpg`
- `progressive_huffman/32x32x8_grayscale_spectral_all_reverse.jpg`

## Differential results

Coefficient references were generated with libjpeg-turbo 3.1.0 built using `WITH_ARITH_DEC=1` and `jpeg_read_coefficients()`.

After the fix:

- All 1,024 quantized coefficients match libjpeg-turbo.
- Coefficients match after every committed progressive scan.
- Direct one-shot and incremental scratch-buffer paths match.
- Strict and non-strict modes match.
- Ascending and descending scan orders produce identical final coefficients.
- Libjpeg-derived coefficient FNV-1a hash:
  `0xa2c3e9111129f380`

Final RGB references were generated using:

`djpeg -rgb -outfile reference.ppm input.jpg`

For both affected fixtures and both strictness modes:

- Maximum channel delta: `1`
- Average channel delta: `0.013671875`

Only one shared 3 KiB RGB oracle is committed because both scan orders produce the same libjpeg-turbo output. Coefficient parity is represented by the hash and selected coefficient assertions rather than additional binary coefficient files.

## Malformed progression policy

Malformed progressive arithmetic files that combine DC and AC in one scan were checked separately.

Existing behavior is unchanged:

- Strict mode rejects them.
- Non-strict mode remains permissive.

This PR does not broaden arithmetic support, weaken progression validation, or address lossless arithmetic JPEGs.

CI is updated to run the optional arithmetic test suite on stable Linux and to compile arithmetic support for the no-std UEFI target.

This work is part of #425


telecos:
A fix to failing Linux Python Wheels CI is proposed in PR #428

PR #429, opened 2026-08-13
Support lossless Huffman images for Chromium parity

telecos:
## Summary

Close a JPEG decoding gap identified during Chromium integration: Chromium's
libjpeg-turbo decoder accepts lossless sequential Huffman JPEGs, while
`zune-jpeg` did not support them.

This adds decoding for 8-bit SOF3 lossless Huffman images with:

- predictors 1 through 7 and point transforms
- interleaved and non-interleaved scans
- row-aligned restart intervals
- exact entropy-boundary handling and recoverable EOF retries
- category 16 differences with 16-bit predictor history

The supported scope currently requires 1x1 component sampling. Unsupported
lossless variants are rejected explicitly.

## Coverage

Coverage uses CC0 JPEG fixtures and sample output generated with
libjpeg-turbo 3.1.0. It verifies:

- pixel parity across all seven predictors
- grayscale and YCbCr output parity
- point transforms and restart markers
- interleaved and non-interleaved scan layouts
- malformed headers, invalid categories, and premature markers
- retries at every entropy-stream byte boundary
- cancellation and repeated `decode_into` behavior

This work is part of #425

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -1335,11 +1335,17 @@ where
                 parse_start_of_frame(marker, self)?;
                 self.is_progressive = is_progressive;
             }
-            Marker::SOF(3) => {
-                trace!("Image encoding scheme =`Lossless Huffman`");
-                parse_start_of_frame(SOFMarkers::LosslessHuffman, self)?;
+            Marker::SOF(3 | 11) => {
+                let (marker, is_arithmetic) = match m {
+                    Marker::SOF(3) => (SOFMarkers::LosslessHuffman, false),
+                    Marker::SOF(11) => (SOFMarkers::LosslessArithmetic, true),
+                    _ => unreachable!()
+                };
+
+                trace!("Image encoding scheme =`{marker:?}`");
+                parse_start_of_frame(marker, self)?;
                 self.is_progressive = false;
-                self.is_arithmetic = false;
+                self.is_arithmetic = is_arithmetic;
             }
             #[cfg(feature = "arith")]
             Marker::SOF(9..=10) => {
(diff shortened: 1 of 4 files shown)


etemesi254:
Hi, thanks for this.

Of concern, does Chromium actually expose these features publicly? 

Like would Chromium decode a lossless image on a normal release build.

Becuase if memory serves me right it does not even decode airthmetic images.

But happy we can decode even more image variants

crates/zune-jpeg/README.md
etemesi254, line 1: If we are putting the details on the readme lets flesh it out from the most common to the least common.

Also we understand most people don't  understand what lossless sequential huffman even stands for so not the best second sentence on a jpeg decoder readme

telecos:
> Hi, thanks for this.
> 
> Of concern, does Chromium actually expose these features publicly?
> 
> Like would Chromium decode a lossless image on a normal release build.
> 
> Becuase if memory serves me right it does not even decode airthmetic images.
> 
> But happy we can decode even more image variants

Thanks for raising this. I checked Chromium's current integration in detail.

Chromium vendors libjpeg-turbo 3.1.0 and compiles lossless Huffman support, including the 8-, 12-, and 16-bit implementations. However, Blink unconditionally requests RGB(A)/BGRA output for grayscale, YCbCr, and RGB
JPEGs. libjpeg-turbo intentionally rejects colorspace-changing output for lossless JPEGs, so a normal Chromium build exposes the header dimensions but fails when pixel decompression starts.

Arithmetic coding is different: Chromium explicitly compiles it out. 

I have narrowed this PR accordingly to match current Chromium behaviour which only is able to parse metadata and provide image size. It now only parses SOF3 headers and exposes dimensions for 2–16-bit lossless Huffman images. Pixel decoding returns the existing unsupported-feature error. 

The support of the full decoding of these type of images would be separated PRs but we will keep them out of scope for now.

crates/zune-jpeg/README.md
telecos: Agreed. Since this PR no longer adds public pixel-decoding support, I removed the README change rather than promoting a niche JPEG process near the top of the decoder overview. A future decoder PR can add a properly ordered capabilities section covering common formats first and specialized variants afterward.

telecos:
Validated in Chromium that these fixes bring corresponding test files to `NoDifference`

https://chromium-review.googlesource.com/c/chromium/src/+/8253433

etemesi254:
Lossless image would also be a great addition, but my suggestion is we have it as another PR when we are done with getting zune-jpeg to Chromium.

But would need to be gated under a feature.

Reviewed this, okay, need your last flag.

(Apologies for delay)

telecos:
> Lossless image would also be a great addition, but my suggestion is we have it as another PR when we are done with getting zune-jpeg to Chromium.
> 
> But would need to be gated under a feature.
> 
> Reviewed this, okay, need your last flag.
> 
> (Apologies for delay)

Thanks for the review. This change is already being used in Chromium as corresponding CL incorporating patch was already completed, so I think this PR can be merged.

PR #432, opened 2026-08-17
JpegDecoder: expose dimensions for 12-bit Huffman images

telecos:
## Summary

Allow 12-bit extended sequential and progressive Huffman JPEG headers to be parsed without enabling 12-bit pixel decoding.

This matches the behavior Chromium observes from its libjpeg-turbo build: image dimensions are available after header parsing, while attempting to decode the pixels still fails as unsupported.

## Changes

- Recognize 12-bit SOF1 and SOF2 frame headers.
- Preserve the correct SOF1 frame classification.
- Report image dimensions and header metadata through the existing APIs.
- Reject 12-bit images before entering the 8-bit pixel decoding path.
- Add tests covering:
  - 12-bit extended sequential header parsing and dimensions.
  - 12-bit progressive header parsing and dimensions.
  - Rejection of invalid 12-bit baseline headers.
  - Continued support for decoding 8-bit SOF1 images.

This does not add 12-bit pixel decoding or change the public API.

This work is part of #425

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -652,6 +652,7 @@ where
     /// See DecodeErrors for an explanation
     pub fn decode(&mut self) -> Result<Vec<u8>, DecodeErrors> {
         self.decode_headers()?;
+        self.ensure_supported_sample_precision()?;
 
         if self.expects_dnl {
             // Height is unknown until DNL is encountered during entropy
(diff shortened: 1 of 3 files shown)


telecos:
Validated in Chromium that these fixes bring corresponding test files to NoDifference

https://chromium-review.googlesource.com/c/chromium/src/+/8253433

PR #435, opened 2026-08-20
Decode MCUs with a block list

197g:
Build a list of blocks in their iteration order, instead of by scan order and deeply nested loops. This mainly helps progressive images due to constant folding (a comment touched on this) but it still simplifies the indexing structure since we no longer have the offset calculation happening during decoding but rather collapse it into a simple `base + stride * j_mcu` that can be updated while iterating.

While implementing this I noticed we're not actually enforcing a section of the specification found in B.2.1 (High-level syntax) where non-DC scans *must not* contain more than one component and thus never should run the mcu-order case. Since that is itself another special case I'm going to investigate if this would benefit from a specialized method or not; and what images may be affected. Maybe some corrupted images actually rely on this being enforced in a particular way.

Opening this as a draft since the main point here is to benchmark, this is a speedup primarily even though it also simplifies logic (in my opinion anyways).

### DIFF
--- a/crates/zune-jpeg/src/decoder.rs
+++ b/crates/zune-jpeg/src/decoder.rs
@@ -29,13 +29,13 @@ use crate::bitstream::{BitstreamStateSnapshot, BitStreamHuffman};
 #[cfg(feature = "arith")]
 use crate::bitstream_arith::{ArithACTables, ArithDCTables, BitStreamArithmetic};
 use crate::color_convert::choose_ycbcr_to_rgb_convert_func;
-use crate::components::{Components, SampleRatios};
+use crate::components::{Components, SampleRatios, ScanBlock};
 use crate::errors::{DecodeErrors, UnsupportedSchemes};
 #[cfg(feature = "arith")]
 use crate::headers::parse_dac;
 use crate::headers::{
     parse_app1, parse_app13, parse_app14, parse_app2, parse_dqt, parse_huffman, parse_sos,
-    parse_start_of_frame, with_marker_body
+    parse_start_of_frame, with_marker_body,
 };
 use crate::huffman::HuffmanTable;
 use crate::idct::{choose_idct_1x1_func, choose_idct_4x4_func, choose_idct_func};
@@ -269,7 +269,14 @@ pub struct JpegDecoder<T> {
     /// Number of components.
     pub(crate) num_scans:        u8,
     /// For a scan, check if any component has vertical/horizontal sampling.
+    ///
+    /// This tells us if blocks occur in the same natural image order in all components part of the
+    /// scan or not. If multiple blocks from a component are part of an MCU they occur in vertical/
+    /// horizontal order of blocks within an MCU, for each MCU.
     pub(crate) scan_subsampled:  bool,
+    pub(crate) scan_blocks:      [ScanBlock; 10],
+    /// Number of component blocks in the scan.
+    pub(crate) num_scan_blocks:  u8,
     // Function pointers, for pointy stuff.
     /// Dequantize and idct function
     // This is determined at runtime which function to run, statically it's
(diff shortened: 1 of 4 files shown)


crates/zune-jpeg/src/headers.rs
```diff
@@ -693,64 +772,67 @@ pub(crate) fn parse_app2<T: ZByteReaderTrait>(
...
+                let body_start_pos = decoder.stream.position()? - body.len() as u64;
+                let mpf_offset = body_start_pos + MPF_DATA.len() as u64;
+                let data = body[MPF_DATA.len()..].to_vec();
+                App2Payload::Mpf {
+                    offset: mpf_offset,
+                    data,
+                }
+            } else {
+                App2Payload::Unknown
+            };
```
197g, line 835: Sorry, I must have hit format-on-save by accident. I can undo this part of the first commit if you wish. It's not a functional change (a few other minor occurrences in this file).

crates/zune-jpeg/src/headers.rs
```diff
@@ -551,23 +557,88 @@ pub(crate) fn parse_sos<T: ZByteReaderTrait>(
...
+        //
+        // This clarifies that for progressive images only the first scan with DC components is
+        // allowed to be interleaved. All other scans with other components are not to be
+        // interleaved, implying those can not contain multiple subsampled components.
+
+        // NOTE: not sure about comparing number of scans to components. It's an internal choice
+        // that we're using a different buffer in these cases but how would that influence the MCU
+        // order of scans that are present? It's not clear from the ITU T.81 specification.
+        let is_image_progressive =
+            usize::from(image.num_scans) != image.components.len();
```
etemesi254, line 576: If this checks on progressive I would suggest we stick to the norm of using the sof flags, its a tricky thing later when one comes to derive this and thinks what it means

crates/zune-jpeg/src/headers.rs
etemesi254: no problem with the format

crates/zune-jpeg/src/mcu.rs
```diff
@@ -584,33 +584,25 @@ impl<T: ZByteReaderTrait> JpegDecoder<T> {
     fn decode_mcu_width<const PROGRESSIVE: bool, B: BitStream>(
```
etemesi254, line 584: Actually on looking into this, the "const PROGRESSIVE" meaning is wrong, had me thinking it was a flag for progressive images, we should change this.

The distinction is those images that require full size buffers instead of widths, but this I can do myself not a major concern

etemesi254:
> Build a list of blocks in their iteration order, instead of by scan order and deeply nested loops. This mainly helps progressive images due to constant folding (a comment touched on this) but it still simplifies the indexing structure since we no longer have the offset calculation happening during decoding but rather collapse it into a simple base + stride * j_mcu that can be updated while iterating.



Do progressive images use that routine? i believe we are talking about "inner_decode_mcu_width" routine here.

crates/zune-jpeg/src/decoder.rs
```diff
@@ -269,7 +271,14 @@ pub struct JpegDecoder<T> {
     /// Number of components.
     pub(crate) num_scans:        u8,
     /// For a scan, check if any component has vertical/horizontal sampling.
+    ///
+    /// This tells us if blocks occur in the same natural image order in all components part of the
+    /// scan or not. If multiple blocks from a component are part of an MCU they occur in vertical/
+    /// horizontal order of blocks within an MCU, for each MCU.
     pub(crate) scan_subsampled:  bool,
+    pub(crate) scan_blocks:      [ScanBlock; 10],
```
etemesi254, line 279: Also why the hardcoded *10*?

197g: Yes, that's from the specification.

197g: See B.2.3. the definition of Csj (scan component selection). It requires:

∑ H_j × V_j ≤ 10

So this is precisely sufficient here.

etemesi254: Could we have it as constant and a bit docs on the reason.

PR #438, opened 2026-08-31
Rewrite progressive AC refinement

197g:
The main cost in its tight loop is zero-tests against spectral coefficients to decide if they will get an additional bit decoded or to use the huffman coded data. Firstly, the loop there was written in an obscure way so I straightened it out in a way that clarifies that each coefficient is visited once.

Secondly, and importantly for performance, we can perform the zero-check far more efficiently with the help of a mask that remembers which coefficients have been initialized already by their position. This avoids countless `i16`-sized reads for a single `u64` register / stack value. The mask is updated when an AC initialization, i.e. the first bit of a magnitude-1 coefficient, is encountered. This also avoids a previously costly (as per comment in code) check for the EOB run portion of each block. The compression switches to a different loop when there are no further coefficients to initialize, which avoids a bunch of instructions. We also want to avoid unecessary refills and other checks and so optimize when only 0s are left which would not consume any bits from the stream. This was previously a rough comparison against the block as UNZIGZAG is costly. Since the mask is in spectral order however we can now do some simple arithmetic.

Result: ~1.7x performance in some progressive benchmarks (on my laptop with slower memory).

-----

I've played around quite a bit with the AC refinement loop during the rewrite, see some intermediate commits, but then converged to roughly a similar structure as on `dev` anyways. Those can be squashed if you'd like.

Also, this was started before the fine restart commits appeared on `dev` and I'm not entirely sure if I'm following the intent of the code here.

### DIFF
--- a/crates/zune-jpeg/src/bitstream.rs
+++ b/crates/zune-jpeg/src/bitstream.rs
@@ -922,21 +922,42 @@ impl BitStream for BitStreamHuffman {
     }
     #[allow(clippy::too_many_lines, clippy::op_ref)]
     fn decode_mcu_ac_refine<T>(
-        &mut self, reader: &mut ZReader<T>, table: &mut HuffmanTable, block: &mut [i16; 64]
+        &mut self, reader: &mut ZReader<T>, table: &mut HuffmanTable, block: &mut [i16; 64],
     ) -> Result<bool, DecodeErrors>
     where
-        T: ZByteReaderTrait
+        T: ZByteReaderTrait,
     {
         let bit = self.successive_low_mask;
+        debug_assert!(bit > 0, "One bit in low mask set");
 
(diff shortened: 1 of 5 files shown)


197g:
The benchmark `DRI / Restart markers/incremental resume` seems to fail for me but also does on #e34d887dd03facceb558c1221ae41bd374181588.


```
thread 'main' (9347) panicked at benchmarks/benches/decode_jpeg.rs:332:27:
scan should not complete at 80% visibility
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I think that's not introduced by my code.

etemesi254:
> Secondly, and importantly for performance, we can perform the zero-check far more efficiently with the help of a mask that remembers which coefficients have been initialized already by their position. 

Quite smart, thank you for this

Issue #442, opened 2026-09-12
zune-jpeg: YCCK images decode to inverted colors, and YCCK -> CMYK is unimplemented


champierre:
## Summary

Decoding an Adobe YCCK JPEG (APP14 transform 2) to `ColorSpace::RGB` returns inverted colors: a light image decodes to nearly black. Asking for `ColorSpace::CMYK` instead fails with `Unimplemented colorspace mapping from YCCK to CMYK`, so there is no way to get correct pixels out.

Reproduced with 0.5.15 and 0.5.16-rc2.

## Reproduction

The image is the **first** `DCTDecode` stream (7200x4500, 10,343,229 bytes, sha256 `77f4994f6346411d23e48923ef1d9932e8e18e970602b8a01f94e9570424744c`) of this PDF:
https://commons.wikimedia.org/wiki/File:Poster_Session.pdf
(CC BY-SA 4.0, Wikimedia Commons)

```rust
let mut h = JpegDecoder::new_with_options(Cursor::new(&data), opts);
h.decode_headers()?;
assert_eq!(h.input_colorspace(), Some(ColorSpace::YCCK));

// RGB: inverted, mean R=34.0 G=25.5 B=26.9 (gray 28.2)
let px = JpegDecoder::new_with_options(Cursor::new(&data),
    opts.jpeg_set_out_colorspace(ColorSpace::RGB)).decode()?;

// CMYK: Err("Unimplemented colorspace mapping from YCCK to CMYK")
```

## Expected

libjpeg (via Pillow) reads the stored components as [171.2, 214.2, 142.8, 209.6]. Adobe stores YCCK inverted, so the CMYK values are [83.8, 40.8, 112.2, 45.4] — a light image, gray 164.8 by the PDF DeviceGray formula. zune returns gray 28.2.

Since this PDF is a single full-page image, the difference is visible by just opening the converted file: the page comes out almost entirely black.

etemesi254:
Hi, thanks for this.

I haven't heard a lot of YCCK sources to test so this is very welcome.

On the change for YCCK to RGB, the code was inspired from https://github.com/nothings/stb/blob/2c980bb59875b0d32144a71867fbdebb2f77cd20/stb_image.h#L3968-L3974

so is it the claim that that is wrong or that we did something wrong before.

PR #443, opened 2026-09-12
zune-jpeg: Add YCCK to CMYK color conversion

champierre:
Fixes the YCCK side of #442.

Decoding a YCCK image to CMYK failed with `Unimplemented colorspace mapping from YCCK to CMYK`, leaving RGB as the only output. That path returns inverted colors for Adobe YCCK, so there was no way to get correct pixels out of these files.

Adobe writes YCCK by inverting CMY and running it through the YCbCr transform, leaving K alone. Undoing that gives back CMYK: convert YCbCr to RGB, invert those three, and copy K. Unlike the YCCK to RGB path this does not multiply by K, because that composites onto white, which a caller asking for CMYK wants to do itself.

The YCbCr conversion runs into a three component scratch buffer because `color_convert_16` only ever writes RGB-sized pixels; asking it for a four component stride interleaves the channels wrongly.

Checked against libjpeg: `test-images/jpeg/four_components.jpg` decodes to CMYK `[240, 58, 21, 216]` with both. The new test in `tests/app14.rs` asserts this.

This does not change the existing `(YCCK, RGB)` output, which is still inverted — that part of #442 needs a separate decision about whether changing it would break current callers.

### DIFF
--- a/crates/zune-jpeg/src/worker.rs
+++ b/crates/zune-jpeg/src/worker.rs
@@ -81,6 +81,15 @@ pub(crate) fn color_convert(
                 output
             );
         }
+        (ColorSpace::YCCK, ColorSpace::CMYK) => {
+            color_convert_ycck_to_cmyk(
+                unprocessed,
+                width,
+                padded_width,
+                color_convert_16,
+                output
+            );
+        }
         (ColorSpace::CMYK, ColorSpace::RGB) => {
             color_convert_cymk_to_rgb::<3>(unprocessed, width, padded_width, output);
         }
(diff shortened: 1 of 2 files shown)


crates/zune-jpeg/src/worker.rs
```diff
@@ -254,6 +263,45 @@ fn color_convert_ycck_to_rgb<const NUM_COMPONENTS: usize>(
...
+///
+/// The YCbCr conversion runs into a three component scratch buffer because `color_convert_16`
+/// only ever writes RGB-sized pixels; asking it for a four component stride would interleave
+/// the channels wrongly.
+#[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+fn color_convert_ycck_to_cmyk(
+    mcu_block: &[&[i16]; MAX_COMPONENTS], width: usize, padded_width: usize,
+    color_convert_16: ColorConvert16Ptr, output: &mut [u8]
+) {
+    let mut rgb = vec![0_u8; width * 3 * (output.len() / (width * 4))];
```
telecos: I believe that this breaks `no_std` build because `vec!` macro is not imported. Could you add `use alloc::vec`?

On performance side, this conversion is called once per output row, so this allocates a new buffer for every scanline. For example, for the 7200×4500 image from #442, that is 4,500 allocations and 97.2 MB of cumulative temporary allocation traffic. Could we reuse a scratch buffer or process chunks through a small stack buffer instead?

etemesi254:
LGTM, but another pr brought merge conflicts.

If we can fix those can be merged

PR #452, opened 2026-09-22
JpegDecoder: Preserve scan rows after marker read-ahead

### LINKED ISSUE #448: [zune-jpeg] Non-interleaved scans lose their final data-unit row when bitstream read-ahead latches the next marker
### Input Image

A minimal 4,130-byte baseline JPEG (`SOF0`, 88x88, 4:2:0) encoded with three single-component scans (`Ns = 1` each) in the order Y, Cb, Cr. The content is an 88x88 checkerboard pattern with rich AC frequency components and high-contrast alternating tiles.

Image

Decoding with the library reproduces the issue:

```sh
cargo run -- -i ~/input.jpg -o ~/output.png
```

### Visual Comparison

| Expected Output (with fix / C++ `libjpeg-turbo`) | Corrupted Output (without fix / `zune-jpeg` 0.5.16) |
| :---: | :---: |
| Image | Image |
| *Correct gray color in bottom rows* | *The bottom 8 rows suffer from zeroed Cb chroma, producing strong discoloration across bottom tiles* |

---

## Issue Description

For a baseline JPEG whose components are coded in separate single-component scans, `zune-jpeg` silently drops the **final data-unit row of the scan that is followed by another scan**. The affected region of that component's plane keeps the value it was allocated with (`0`), not the neutral `128`, because the IDCT never runs over it and so the level shift is never applied.

With scan order Y, Cb, Cr the victim is Cb. On the alternating grayscale tiles (e.g. `RGB(53, 53, 53)` and `RGB(98, 98, 98)` where `Cb = Cr = 128`), rows 80–87 decode as `RGB(53, 97, 0)` and `RGB(98, 142, 0)`: red is untouched, green is elevated by ~44 (`-0.344 * -128`), and blue clamps to 0 (`Y - 227 < 0`), which is exactly what an uninitialized `Cb = 0` produces. Row 79 is collateral damage from vertical chroma upsampling blending the last good chroma row with the zeroed one.

The output is a valid, plausible-looking image, so this fails silently.

### Alignment Trigger

This failure is triggered by bitstream refill alignment rather than image dimensions. Holding the image at 88x88 and sweeping only the quality setting:

| quality | 20 | 30 | 40 | 50 | 60 | 70 | 75 | 80 | 85 | 90 | 92 | 95 | 98 |
| :-- | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | 
| result | bad | bad | bad | bad | ok | ok | ok | bad | bad | bad | ok | bad | ok |

Same dimensions, same sampling factors, same scan structure; only the number of compressed bytes changes. What decides the outcome is whether the four-byte read-ahead crosses into the next marker before or after the last row is decoded.

### Why standard interleaved JPEGs are unaffected

In standard interleaved JPEGs, the entire image is encoded in a single scan containing all components (`Ns = 3`), terminated by `EOI`. In `check_stream_marker_after_mcu_width`, `Marker::EOI` is acknowledged and ignored (`stream.marker().take()`), returning `Ok(McuContinuation::Ok)` rather than aborting the scan. The row loop therefore always runs to completion. Scan truncation can only occur when a scan is followed by another scan or inter-scan marker (`SOS`, `DHT`, etc.).

---

## Root Cause Analysis

The bitstream refill reads ahead, and the row-end marker check treats anything it finds as the end of the scan.

1. **`refill()` latches a marker while usable bits remain**: In `crates/zune-jpeg/src/bitstream.rs`, `refill()` tops the bit buffer up four bytes at a time. When it encounters `0xFF` followed by a non-zero marker byte, it records `self.marker = Some(marker)` while `bits_left > 0`. The remaining bits belong to the current scan and still need to be decoded.
2. **The row-end check acts on that marker immediately**: `check_stream_marker_after_mcu_width` runs at the end of every data-unit row. On seeing a latched `SOS` or inter-scan marker (`DHT`, `DQT`, `APP`), it immediately returns an early continuation (`AnotherSos` or `InterScanMarker`).
3. **The caller abandons the rest of the scan**: In `decode_mcu_ycbcr_baseline_inner`, the dispatch is `McuContinuation::AnotherSos => continue 'sos`, which restarts the outer scan loop and discards whatever rows the current scan had left.

The mid-row check earlier in the function correctly waits for `stream.bits_left() == 0`, but the row-end check had no equivalent guard.

---

## Proposed Fix

Thread `is_last_row: i + 1 >= scan_du_height` down to the marker check, and ignore scan-terminating markers until the scan's final row:

```diff
diff --git a/crates/zune-jpeg/src/mcu.rs b/crates/zune-jpeg/src/mcu.rs
--- a/crates/zune-jpeg/src/mcu.rs
+++ b/crates/zune-jpeg/src/mcu.rs
@@ -42,6 +42,8 @@ struct McuWidthContext<'a, B: BitStream>
     stream: &'a mut B,
     // Full-image coefficient buffers for multi-SOS baseline scans.
     progressive: &'a mut [Vec<i16>; MAX_COMPONENTS],
+    // True if this is the final data-unit row of the current scan.
+    is_last_row: bool,
 }
 
 impl<T: ZByteReaderTrait> JpegDecoder<T> {
@@ -340,6 +342,7 @@ impl<T: ZByteReaderTrait> JpegDecoder<T>
                         tmp: &mut tmp,
                         stream: &mut stream,
                         progressive: &mut *progressive_mcus,
+                        is_last_row: i + 1 >= scan_du_height,
                     };
                     if all_components_in_first_scan {
                         self.decode_mcu_width::<false, B>(&mut mcu_width_context)
@@ -679,6 +682,7 @@ impl<T: ZByteReaderTrait> JpegDecoder<T>
         let tmp: &mut [i32; 64] = &mut *context.tmp;
         let stream: &mut B = &mut *context.stream;
         let progressive: &mut [Vec<i16>; MAX_COMPONENTS] = &mut *context.progressive;
+        let is_last_row = context.is_last_row;
 
         let z_order = self.z_order;
         let z_scans = &z_order[..usize::from(self.num_scans)];
@@ -874,11 +878,11 @@ impl<T: ZByteReaderTrait> JpegDecoder<T>
             }
         }
 
-        self.check_stream_marker_after_mcu_width(stream)
+        self.check_stream_marker_after_mcu_width(stream, is_last_row)
     }
 
     fn check_stream_marker_after_mcu_width<B: BitStream>(
-        &mut self, stream: &mut B,
+        &mut self, stream: &mut B, is_last_row: bool,
     ) -> Result<McuContinuation, DecodeErrors> {
         // After all interleaved components, that's an MCU
         // handle stream markers
@@ -891,6 +895,20 @@ impl<T: ZByteReaderTrait> JpegDecoder<T>
         // But libjpeg-turbo allows it because of some weird reason. so I'll also
         // allow it because of some weird reason.
         if let Some(m) = stream.marker() {
+            // Suppress scan-terminating markers until the scan's final row is decoded,
+            // as bitstream refill read-ahead may latch them while rows remain.
+            let ends_scan = matches!(
+                m,
+                Marker::SOS
+                    | Marker::DAC
+                    | Marker::DHT
+                    | Marker::DQT
+                    | Marker::DRI
+                    | Marker::COM
+                    | Marker::APP(_)
+            );
+            if ends_scan && !is_last_row {
+                return Ok(McuContinuation::Ok);
             }
             if *m == Marker::EOI {
                 // acknowledge and ignore EOI marker.
                 stream.marker().take();
```

telecos:
## Summary

Fix baseline non-interleaved JPEG decoding when bitstream read-ahead discovers the next scan marker before all expected data-unit rows have been processed.

## Root cause

The bitstream refiller can latch the next `SOS`, `DHT`, or other inter-scan marker while usable entropy bits remain buffered.

The baseline row-end marker check treated the latched marker as immediately authoritative and advanced to the next scan. This could leave the final data-unit row of the current component undecoded.

## Fix

Track whether the decoder is processing the final expected row of the current scan.

When read-ahead has latched a scan-ending marker while buffered entropy data and expected rows remain, retain the marker and continue decoding. Existing handling remains unchanged for `EOI`, restart markers, and `DNL`.

Fixes #448


etemesi254:
Sorry, some MR, could we sort that out then I can merge it.

Really appreciate the work put in.

telecos:
I resolved merge conflict

